PR #19, opened 2020-02-19
Update README.md

bouzaghrane:
Updated the project structure graph.

@timothyb0912 Do you think I should add a subfolder under `\references\` for reading materials or should we leave it as is?


timothyb0912:
I've got no real opinion. I'll follow your lead with whatever structure you go with!

Issue #15, opened 2020-02-19
Helpful readings


timothyb0912:
Hey everyone,

I'm starting this issue to collect information about readings that we believe to be extremely relevant to our work for the conference in June.

Here are three documents that should be helpful.

I'll add them to the references directory.

First, "Addressing Endogeneity in Discrete Choice Models: Assessing Control-Function and Latent-Variable Models" by Guevara and Ben-Akiva (2010). This is essentially the discrete choice modeller's approach to dealing with unobserved confounding. I think that the main difference we'll be able to show is in how we include the latent confounders.

Second, here are two papers that aside from the Deconfounder work of Blei et al, demonstrate *the same* technique of using latent variable models to model the unobserved confounders.

"Bayesian Sensitivity Analysis for Offline Policy Evaluation" by Jung, Shroff, Feller, and Goel (2019). Note that Avi Feller is at Berkeley, and since his paper is so extremely relevant to our work, we should invite him to review our work before the conference! He'll likely be able to provide great and essential input. Also, I've met and spoke with Avi a couple of times before: he's was very nice in my opinion!

"Assessing Sensitivity to Unmeasured Confounding Using a Simulated Potential Confounder" by Carnegie, Harada, and Hill (2016).

timothyb0912:
@bouzaghrane @hassanobeid1994

timothyb0912:
I added three more papers.

One is the model criticism for causal inference paper by Tran et al.

Another is "Deconfounding Reinforcement Learning in Observational Settings" by Lu, Scholkopf, and Hernandez-Lobato. On page 3, they write

> "To this end, we present a general formulation for addressing RL problems with observational data, namely deconfounding reinforcement learning (DRL). More specifically, given several common confounding assumptions, we first estimate a latent-variable model from observational data. Under some conditions for identification, through the latent-variable model we can simultaneously discover the latent confounders and infer how they affect actions and reward. Then the confounders in the model can be adjusted for, using the causal language developed by Pearl (2009), and finally we optimize the policy based on the resulting deconfounding model."

Lastly, there is the "Causal Effect Inference with Deep Latent-Variable Models" paper by Louizos, Shalit, Welling, et al. (2017) where they "build on recent advances in latent variable modeling to simultaneously estimate the unknown latent space summarizing the confounders and the causal effect". This seems to really mark the start of a bunch of work in this vein from ML researchers.

PR #20, opened 2020-02-20
Reorganizing folder and adding two more papers.

timothyb0912:
Related to Issue #15


timothyb0912:
@bouzaghrane let me know if this addition of a sub-folder for papers looks alright to you!

Also added two more papers, this time from ML researchers using latent-variable models to account for unobserved confounding.

bouzaghrane:
Good with all changes, will be making another change now by adding a new subfolder for reference code

PR #23, opened 2020-02-22
Moved jupyter notebook from references to new code subfolder

bouzaghrane:
created a new subfolder for reference code and moved exiting code reference


bouzaghrane:
It's fair to delete this branch now.

PR #24, opened 2020-02-24
initializing causal graph

hassan-obeid:


timothyb0912 (changes requested):
The predictive checks for the fitted factor model are crucial and really MUST be addressed.

There's no point in interpreting a model that could be a terrible descriptor of our data.

notebooks/working/Causal_Graph_Tim_Data.py
```diff
@@ -0,0 +1,446 @@
...
+        
+        
+    ]
+)
+
+# draw return a graphviz `dot` object, which jupyter can render
+sprinkler.draw()
+# -
+
+# # MNL specification
```
timothyb0912, line 82: Currently, you go directly from the creation of a causal graph (i.e. step 2 in section 5.2 of the causal inference paper I have with Joan), to MNL estimation (i.e step 5).

So far step 3 (attempted falsification of the causal graph) and step 4 (identification of the causal effect) are missing.

Please include code here for those two steps.

notebooks/working/Causal_Graph_Tim_Data.py
```diff
@@ -0,0 +1,446 @@
...
+import pylogit as cm
+from functools import reduce
+
+
+
+import os
+os.listdir('.')
+# -
+
+data = pd.read_csv('spring_2016_all_bay_area_long_format_plus_cross_bay_col.csv')
```
timothyb0912, line 32: Adjust the data path to read from `data` directory (so the notebook can actually be executed without error).

timothyb0912 (changes requested):

timothyb0912 (approved):

Issue #28, opened 2020-02-25
Retrospectives


timothyb0912:
Hey @bouzaghrane and @hassanobeid1994:

I've forgotten about this until now, but I think it's a good idea for us to set aside some time (e.g. 10 min) each week to be thoughtful and reflect on how the project is going.

To that end, I'm starting a retrospectives folder with a markdown document for each of us.

The hope is that before our weekly meetings, we'll each take ten minutes to answer:
- what is one thing that went well this week?
- what is one thing that went poorly this week?
- what is one specific way to improve our project execution process in the next week?
- what is one thing we learned this week?

timothyb0912:
How do you all feel about this?

bouzaghrane:
I agree on this.

PR #29, opened 2020-02-25
Initializing files for retrospectives.

timothyb0912:
Addresses Issue #28 and #33.


timothyb0912:
@bouzaghrane and @hassanobeid1994:

Let me know if you two are okay with the idea of doing weekly retrospectives (10 min/wk)!

PR #30, opened 2020-02-25
Adding literature review of how one can falsify one's causal graph.

timothyb0912:
Opening a pull request to store my literature review on work that has been done about how to falsify a given causal graph. Addresses #33.


timothyb0912:
@hassanobeid1994 since this is directly related to the comment I left on your PR.

Also tagging @bouzaghrane since you should likely do similar checks (if only to make sure everything is working as expected and to see what happens with data not only theory) for your simulation.

The idea is not that we'd do all of these, but simply to take note of all or most of what has been done so we can choose from the list.

Issue #31, opened 2020-02-26
Refactoring Guidance


timothyb0912:
Hey @bouzaghrane and @hassanobeid1994!

I realized that I asked you two to refactor your simulation notebooks so that they use functions and classes but that I have not been very specific.

I'm opening this issue to track the fact that I'm going to add a quick guide detailing my suggestions for how to refactor your notebooks!

I'll attempt to be more detailed and paint a clearer vision of what I think we should aim for with our refactoring!

PR #32, opened 2020-02-26
Creating refactoring guide.

timothyb0912:
Addresses issue #31 and #33. It provides a guide with my suggestions for how to refactor the project's jupyter notebooks.

### DIFF
--- a/references/refactoring_guide.md
+++ b/references/refactoring_guide.md
@@ -1,0 +1,67 @@
+# Refactoring jupyter notebooks: End Goal
+In my opinion, a Jupyter notebook should do something.
+
+Ideally, each notebook demonstrates one or more steps of the project workflow.
+
+In this sense, we should imagine Jupyter notebooks to be [integration tests](https://www.fullstackpython.com/integration-testing.html).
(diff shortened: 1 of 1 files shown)


timothyb0912:
@bouzaghrane and @hassanobeid1994 let me know if you have:
1. questions about anything in the guide
2. comments on anything I wrote, suggestions for improving the guide, etc.
3. disagreements with any refactoring steps or anything I've written
4. no comments but are generally good with using the guide to direct your notebook refactoring.

bouzaghrane:
Thanks @timothyb0912 . Here is a [video tutorial](https://www.youtube.com/watch?v=G1lDk_WKXvY&t=1576s) for how we can use different code quality packages to make sure we're following the guidelines. Each package follows the pep8 rules to a certain extent and therefore I think it probably won't hurt (even though might not be necessary) to run our code through a series of these packages in a set order. Let me know what you all think.

timothyb0912:
@bouzaghrane I agree with it not hurting to run the code through a series of the packages in a certain order. There's a concept of git hooks that executes a given set of commands on certain file types whenever you check them in. I would start by looking into that (see a python flake8 tutorial [here](https://ljvmiranda921.github.io/notebook/2018/06/21/precommits-using-black-and-flake8/)) for setting something like this up!

PR #37, opened 2020-03-02
Demo cit

timothyb0912:
(Partially) addresses issue #36 .

Please take a look: @hassanobeid1994 and @bouzaghrane. 

Essentially, this is what I imagine row 1 of the [refactoring guide](https://github.com/hassanobeid1994/tr_b_causal_2020/blob/master/references/refactoring_guide.md) in the "evolving notebook states" section to look like.

The first portion of the notebook contains the various function definitions and logic needed to get the prototype working. The second portion of the notebook contains a visual test of the functionality and logic contained in the notebook.

In particular, there are two things to pay attention to. One, there is is both a test that is visually performed and a written rationale of the logic behind the test. Two, if you look closely in the `_make_regressor` function, you'll see commented out lines that detail a version of the function that failed the visual test. This information is left in the function on purpose so that later on, during refactoring, this information can be captured in a regression test.

Please follow a similar pattern in your prototypes, and suggest improvements to this if you have any! And as usual, let me know if there are questions about anything I did.

PR #40, opened 2020-03-04
adding first draft of project plan

hassan-obeid:
I slightly modified the approach for the deconfounder by adding a step that is also based on the simulated data. Curious about your thoughts.


references/organization/_06-Project-Plan-and-Workflow.md
```diff
@@ -0,0 +1,64 @@
...
+    - One based on a realistic causal graph with confounders. Preferably focus on a case with one latent confounder (like distance? Age? Gender?) <br>
+-	For each of the two causal graphs above, simulate choice data based on the outcome model assumed. <br>
+-	Estimate the choice model for each of those two datasets. <br>
+-	Repeat the process n times, and calculate the percentage of those times where the estimated parameters’ confidence intervals contain the true parameter. <br>
+
+### Dealing with latent confounders in the causal graph
+
+In this part of the work, we take the realistic causal graph from part 1 and remove a latent confounder from it. We then attempt to recover the latent confounder using the de-confounder algorithm proposed by Blei et al. (2018). We will apply this work to the simulated data set, where we know the true parameters of interest. We will repeat the same simulation process in part 1 where we sample from the population and estimate models where we control for the latent confounder using the deconfounder approach and look at the percent of time we recover the true parameter.
```
timothyb0912: nit: "[...] remove a confounder from the dataset, making the confounder latent with respect to the graph."

references/organization/_06-Project-Plan-and-Workflow.md
```diff
@@ -0,0 +1,64 @@
...
+-	Repeat the process n times, and calculate the percentage of those times where the estimated parameters’ confidence intervals contain the true parameter. <br>
+
+### Dealing with latent confounders in the causal graph
+
+In this part of the work, we take the realistic causal graph from part 1 and remove a latent confounder from it. We then attempt to recover the latent confounder using the de-confounder algorithm proposed by Blei et al. (2018). We will apply this work to the simulated data set, where we know the true parameters of interest. We will repeat the same simulation process in part 1 where we sample from the population and estimate models where we control for the latent confounder using the deconfounder approach and look at the percent of time we recover the true parameter. 
+
+##### Concrete steps: 
+
+Functions <br>
+-	A class with functions to estimate factor models. Start with PPCA as in the deconfounder’s tutorial and expand as needed. <br>
```
timothyb0912, line 50: I would explicitly add that the class needs to have (separate) methods that allows it to:
- sample from the posterior of the latent confounders given the X's
- return the expected value of the latent confounders given the X's

references/organization/_06-Project-Plan-and-Workflow.md
```diff
@@ -0,0 +1,64 @@
...
+-	Then, come up with a “utility level” causal graph, where we come up with a causal graph for each utility equation since the confounding is at the level of the utilities in a MNL context. These causal graphs would inform the utility specifications, and will determine which utilities do and do not require estimating a latent confounder. We discussed this briefly before, but are we on the same page? <br>
+-	Based on the above specifications, estimate the required latent confounders using a factor model, start with PPCA. <br>
+-	Perform posterior predictive checks on the estimated factor model and assess goodness of fit. Since we’re doing this on a simulated dataset, then we know in advance how many latent confounders there are. We could potentially try controlling for more confounders and see how that affects the results (in other words, see how sensitive our analysis is to the number of confounders used). <br>
+-	Once the checks pass, include the latent confounder in the utility equations, and estimate a MNL. <br>
+-	Repeat the process n times, and calculate the percentage of those times where the estimated parameters’ confidence intervals contain the true parameter. <br>
```
timothyb0912: I would only do the visual posterior predictive checking once (as opposed to n times). These take some time to interpret visually and you def won't be doing 100s of these.

You should leaving numeric posterior predictive checks in the code though. The idea would be to have a statement like `if not min_p_value < posterior_predictive_p_value and posterior_predictive_p_value < max_p_value: raise ModelValidationError`.

references/organization/_06-Project-Plan-and-Workflow.md
```diff
@@ -0,0 +1,64 @@
...
+Finally, we can repeat the above process, but now on the full, real bike dataset. The workflow would be the same as above, with the addition of one step: <br>
+-	After hypothesizing on a causal graph, we need to do some conditional independence testing to falsify the graph by making sure nothing clearly violates conditional independence tests. The caveat here is that there don’t seem to be one way to perform conditional independence testing – there are many ways proposed, but I’m not sure if the problem itself is solved yet. We can try one way, and point out that this is an active area of research and that there exists other ways to do it. The important take away is that the causal graph should not clearly violate a given test, in which case something is obviously wrong. <br>
```
timothyb0912, line 63: I think we should call out two explicit demonstrations here:
1. a test of conditional independence amongst the observable variables in the graph
2. a test of the implied relationships due to the latent variables and their hypothesized structure

For the first one, you've got a demonstration in #37 .
For the second, I will brainstorm!

Either way, everything will be couched exactly as you said: many ways proposed, but no matter which we use, we should not clearly violate any given test.

references/organization/_01-ProblemStatement.ipynb
```diff
@@ -11,14 +11,9 @@
...
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "This project attempts to bridge the gap between causal inference modeling practices and the field of demand modeling (discrete choice modeling) in transportation by demonstrating an approach to think causally about the modeling problem."
+    "## What is the problem motivating this project?\n",
+    "\n",
+    "The existing disconnect between the fields of travel demand modeling and causal inference make it hard for cross-fertilization between the two fields to take place. This project is motivated by the current lack of use of methods and findings from causal inference in travel demand modeling."
```
timothyb0912, line 16: 👍

references/organization/_06-Project-Plan-and-Workflow.md
```diff
@@ -0,0 +1,64 @@
...
+-	Based on the above specifications, estimate the required latent confounders using a factor model, start with PPCA. <br>
+-	Perform posterior predictive checks on the estimated factor model and assess goodness of fit. Since we’re doing this on a simulated dataset, then we know in advance how many latent confounders there are. We could potentially try controlling for more confounders and see how that affects the results (in other words, see how sensitive our analysis is to the number of confounders used). <br>
+-	Once the checks pass, include the latent confounder in the utility equations, and estimate a MNL. <br>
+-	Repeat the process n times, and calculate the percentage of those times where the estimated parameters’ confidence intervals contain the true parameter. <br>
+Finally, we can repeat the above process, but now on the full, real bike dataset. The workflow would be the same as above, with the addition of one step: <br>
```
timothyb0912: Addition of three steps actually. Once one has tried to falsify the causal graph and fit the model to the entire dataset (aka including the outcome model, i.e. the MNL with the latent confounders), one should check:
- the outcome model and
- the outcome + treatment-assignment-model (aka the model of the X's)

Luckily, both of these two are essentially just applying the techniques from my model checking paper and we can reuse the code from that project.

timothyb0912 (changes requested):
I gave it a look over and the first draft of the goals of what need to be done all look good to me!

See my line-by-line comments for specifics.

As a next step, after addressing the comments mentioned above, add estimates of how long each of the steps in the workflow should take.

Specifically, for each part of the project (again thinking 3 parts), imagine that there is a total of 40hrs to be allocated. Actually put the amount of hours next to each bullet point under the "Concrete Steps" headings, and ensure that each of the tasks looks like it can be reasonably done in 40hrs (aka 4 weeks).

references/organization/_06-Project-Plan-and-Workflow.md
```diff
@@ -0,0 +1,64 @@
...
+-	Since we’re dealing with a multinomial choice problem, we need a function to simulate outcomes and be flexible to take in multiple specifications for each utility equation.
+
+Workflow <br>
+-	Come up with an outcome (choice) model based on the bike data in Brathwaite (2018). The model doesn’t need to be the exact model estimated in the paper, but that can be the case. We just need to select a set of variables while being mindful that one of those variables will be used as a latent confounder in the next step. 
+-	Simulate two sets of data for the X variables: <br>
+    - One where all the Xs are independent (i.e. the only edges in the causal graph are between the X and the outcome variable). 
+    - One based on a realistic causal graph with confounders. Preferably focus on a case with one latent confounder (like distance? Age? Gender?) <br>
+-	For each of the two causal graphs above, simulate choice data based on the outcome model assumed. <br>
+-	Estimate the choice model for each of those two datasets. <br>
+-	Repeat the process n times, and calculate the percentage of those times where the estimated parameters’ confidence intervals contain the true parameter. <br>
```
bouzaghrane: ..contain the true parameters from the assumed outcome choice model.

references/organization/_06-Project-Plan-and-Workflow.md
```diff
@@ -0,0 +1,64 @@
...
+-	Since we’re dealing with a multinomial choice problem, we need a function to simulate outcomes and be flexible to take in multiple specifications for each utility equation.
+
+Workflow <br>
+-	Come up with an outcome (choice) model based on the bike data in Brathwaite (2018). The model doesn’t need to be the exact model estimated in the paper, but that can be the case. We just need to select a set of variables while being mindful that one of those variables will be used as a latent confounder in the next step. 
+-	Simulate two sets of data for the X variables: <br>
+    - One where all the Xs are independent (i.e. the only edges in the causal graph are between the X and the outcome variable). 
+    - One based on a realistic causal graph with confounders. Preferably focus on a case with one latent confounder (like distance? Age? Gender?) <br>
+-	For each of the two causal graphs above, simulate choice data based on the outcome model assumed. <br>
+-	Estimate the choice model for each of those two datasets. <br>
```
bouzaghrane: Estimate the choice model for each of the two datasets using the assumed choice model specification.

references/organization/_06-Project-Plan-and-Workflow.md
hassan-obeid: @timothyb0912  While I agree on this, I don't necessarily 100% agree that it is necessary for the purposes of our work. Meaning -- if our models don't pass your checks, that conclusion wouldn't have a direct implication on the comparison of our approach vs. the no-causal graph/no latent confounder approach. I think this should be done for the eventual paper, but it's not directly relevant to the presentation.

references/organization/_06-Project-Plan-and-Workflow.md
hassan-obeid: @timothyb0912  I don't follow you on the second test. We can discuss more later, but I'll leave it out for now.

timothyb0912: Yeah no worries. I'm happy to discuss if you want but I second your suggestion of leaving it out for now as these sort of tests look difficult/non-trivial to implement based on my readings!

references/organization/_06-Project-Plan-and-Workflow.md
timothyb0912: I'm with you, partially.

If we only talk about the simulated results, then I think you're right: the posterior predictive checks aren't **necessary** as we know we're simulating-from / estimating the data-generating-model.

If we speak about the real data though, I think the posterior predictive checks are necessary.

Assuming we find differences (as your prototype suggests we will) between the deconfounder based choice model, and the standard choice model, we still will not know whether or not we should put much stock in the deconfounder results.

In other words, our justification for drawing conclusions based on the deconfounder with real data is based on our model being a good fit to the real data, so I think we need to provide an abundance of evidence for the adequacy of the model fit (as with posterior predictive checks).

timothyb0912 (commented):
These changes look good so far to me! Keep them coming!

PR #51, opened 2020-03-11
Ho simulation simple scenario

hassan-obeid:
Hi folks, 

Please take a look at this notebook and let me know what you think. It affects our story for the first part of the work. This is what I wanted to discuss with you all ideally on a call before Sunday, if we can't discuss this by email.


bouzaghrane:
@hassanobeid1994 can you add a jupytext integration to your notebooks and linking it to a markdown file so it can be easy to add comments to specific lines of code? Not a priority but can be done after today's call

notebooks/working/Simple_Simulation.md
```diff
@@ -0,0 +1,382 @@
...
+time_on_dist_std = time_on_distance_reg.resid.std()
+cost_on_dist_std = cost_on_distance_reg.resid.std()
+
+sample_size = len(drive_alone_df)
+```
+
+```python
+total_travel_time_sim = ( time_on_distance_reg.params[0] + 
+                         time_on_distance_reg.params[1]*drive_alone_df['total_travel_distance']
+                        + np.random.normal(loc=0, scale=time_on_dist_std, size = sample_size) )
```
timothyb0912, line 126: Before simulating the various variables, I would suggest adding a call to `np.random.seed` to make the results of the simulations reproducible.

notebooks/working/Simple_Simulation.md
```diff
@@ -0,0 +1,382 @@
...
+utilities = asc_drive + np.dot( simulation_df, coeffs) + np.random.normal(loc=0, scale = 1, size = sample_size)
+
+simulation_df['utilities'] = utilities
+```
+
+## Estimation
+
+Note that here, I'm treating the utilities as an observed quantity that I'm trying to estimate. This will get more complicated as we include different modes and actually try to maximize the correct likelihood function. 
+
+The thing I need to point out here is that irrespective of our causal graph, we will always recover the paramters in coeffs (defined above) when we run the regression. The question becomes, are the coefficients on each of the variables (the true coefficients) causal? That depends on the causal graph:
```
timothyb0912, line 163: > [...] irrespective of our causal graph, we will always recover the paramters in coeffs (defined above) when we run the regression.

Why is this believed to be true?

hassan-obeid: @timothyb0912 based on what I've done so far and my knowledge of things. I know you have an idea/counter-example in mind, so I want to discuss that with you.

notebooks/working/Simple_Simulation.md
```diff
@@ -0,0 +1,382 @@
...
+    return results
+```
+
+```python
+## Regress travel cost on travel distance
+
+cost_on_distance_reg = fit_regression(X=['total_travel_distance'], 
+                                      y = ['total_travel_cost'], data=drive_alone_df, plotting = True)
+
+cost_on_distance_reg.summary()
```
timothyb0912, line 101: Hmm, from the plots travel cost looks categorical (hence the distinct diagonal lines in the plot of the residuals vs travel distance. You could confirm by checking out `drive_alone_df['total_travel_cost'].value_counts()`

Perhaps you should use an MNL instead of a linear regression for predicting travel distance?

notebooks/working/Simple_Simulation.md
```diff
@@ -0,0 +1,382 @@
...
+
+The thing I need to point out here is that irrespective of our causal graph, we will always recover the paramters in coeffs (defined above) when we run the regression. The question becomes, are the coefficients on each of the variables (the true coefficients) causal? That depends on the causal graph:
+
+- In the case of independent variables, the true causal effect of distance is the 0.1, the same value in the coeffs array. Thus, running a regression on all the variables in this case would return the true causal estimate. 
+
+- In the case where travel cost and travel time are descendents of distance, the true causal effect of distance becomes: 0.1 - 0.5*(1.28) - 1.5*(0.22) = -0.87. We will only recover this value if we omit travel distance and travel time from the utility equation. Alternatively, we can keep them in the equation, but then don't assign the coefficient for distance any causal interpretation, and post-process the results (and make parametric assumptions!) to calculate the true causal effect.
```
timothyb0912, line 167: These numbers no longer match the example shown in the code though I imagine they once did. I think the needed edit is
```suggestion
- In the case where travel cost and travel time are descendents of distance, the true causal effect of distance becomes: `3 - 0.5*(1.28) - 1.5*(0.22) = 2.03`. We will only recover this value if we omit travel distance and travel time from the utility equation. Alternatively, we can keep them in the equation, but then don't assign the coefficient for distance any causal interpretation, and post-process the results (and make parametric assumptions!) to calculate the true causal effect. 
```

notebooks/working/Simple_Simulation.md
```diff
@@ -0,0 +1,382 @@
...
+
+# X_raw = np.array([s2,s3]).reshape((1000,2))
+# X =( X_raw - X_raw.mean(axis=0) )/X_raw.std(axis=0)
+
+
+U, Vt, train_err, test_err, M, resid = cv_pca(data=X, rank=2)
+train_err, test_err
+```
+
+### Check PCA
```
timothyb0912, line 282: To check one's PCA (and to check one's models in general), one suggestion I have is to try to always include at least one check that is essentially just a plot of the original data vs the simulated data.

So for the PCA, I might (for each variable) do something like plot the simulated vs observed distributions. For e.g. try looking at a kernel density estimate (or empirical, cumulative distribution function) of the PCA-simulated travel costs vs the observed travel costs.

Issue #47, opened 2020-03-09
[TB] Week of Mar 8


timothyb0912:
- [x] Investigate why travel distance is zero for all transit modes.
  - Deliverable: Pushed markdown file noting the data issue (and reason if found) for future users of the data.
  - Estimated: 3 hours.
  - Actual: 1 hour.
- [x] Review Amine's revised distribution fitting code based on long-format data
  - Deliverable: Comments on the updated PR (#44)
  - Estimated: 2 hr.
  - Actual: 4 hrs.
- [x] Review Hassan's code to simulate data for a single, latently confounded utility equation.
  - Estimated: 2 hr.
  - Actual: 2. hrs.
- [x] Weekly meeting.
  - Estimated: 2 hr.
  - Actual:

bouzaghrane:
@timothyb0912 I just wanted to check in with you about the status of task one in your list. Any luck?

timothyb0912:
@bouzaghrane I have not looked a single second at it. It's my midweek task, so I'll have it done by tomorrow morning and tag you on it.

PR #52, opened 2020-03-12
Adding readme with data notes.

timothyb0912:
Addresses task 1 of #47

### DIFF
--- a/data/README.md
+++ b/data/README.md
@@ -1,0 +1,16 @@
+# Description
+The dataset used in this project is taken from Brathwaite and Walker (2018).
+```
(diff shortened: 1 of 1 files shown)


timothyb0912:
@bouzaghrane and @hassanobeid1994 
This is why the transit distances are zero.

The MTC travel skims don't define a distance between OD pairs for transit.
For instance, you need to know the route and transit mode someone took to know the distance. Which bus route? Bus or Train? etc.

bouzaghrane:
@timothyb0912 sounds good to me.

PR #56, opened 2020-03-16
Update falsification review 1

timothyb0912:
Updating falsification guide with references to trek separation, how its depicted graphically, and how we could use it to falsify our hypothesized causal graphs with latent confounders.


timothyb0912:
Tagging you all for visibility in case you're interested!
@bouzaghrane 
@hassanobeid1994

Issue #60, opened 2020-03-22
[TB] Week of March 22


bouzaghrane:
- [x] Draft email to Blei and send to Amine and Hassan:
  - Deliverable: Email to Amine and Hassan
  - Estimate: 1 hr.
  - Actual: 1 hr.

- [x] Email Amine the code for simulating from model in pylogit:
  - Deliverable: pushed code to reference folder in separate branch
  - Estimate: 1 hr.
  - Actual: 1 hr.

- [x] Prototype the use of marginal independence tests for falsifying one's causal graph. Priority number 3. Carryover task from last week.
   - Deliverable: Environment.yml file storing instructions for reproducing the conda environment I'll use for this task.
   - Deliverable: Pull-request with notebook and accompanying .md / .py file showing marginal independence tests for the causal graph in our example.
   - Estimated: 2 hrs.
   - Actual: 1.5 hrs

- [x] Start branch and pull-request for prior and posterior predictive tests of the completed data independence assertions from the deconfounder's causal graph.
  - Deliverable: new branch and pull-request showing notebook with pseudocode for the predictive tests.
   - Deliverable: Updated github issues and pull-request comments to reflect the new PR.
  - Estimated: 3 hrs
  - Actual: 3hrs

- [x] Review Amine and Hassans' work.
   - Deliverable: Comments on Amine and Hassan's respective pull-requests.
   - Estimated: 3 hrs.
   - Actual: 5hrs.

- [x] Weekly Meeting:
   -  Estimate: 2 hours
   - Actual:

bouzaghrane:
@timothyb0912 feel free to add estimates in there and add anything if I missed it.

timothyb0912:
@bouzaghrane for code showing simulation from a pylogit model, see cell 7 here: https://github.com/adb40/machine_learning_session/blob/TimB-files/src/scripts/_03_tb-MNL_Checking.ipynb

You can grab the needed functions from here: https://github.com/adb40/machine_learning_session/tree/TimB-files/src/scripts/viz

and in particular from the function here: https://github.com/adb40/machine_learning_session/blob/TimB-files/src/scripts/viz/utils.py#L225

PR #65, opened 2020-03-25
Adding viz module.

timothyb0912:
Adding the pylogit.viz module to the project so that we can have access to its function for simulating choices from a given matrix of simulated long-format probabilities.

Addresses task 2 of #60.

### DIFF
--- a/src/viz/__init__.py
+++ b/src/viz/__init__.py
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+"""
+Visualization module for PyLogit.
+"""
+from __future__ import absolute_import
+
+from .utils import (progress,
+                    is_categorical,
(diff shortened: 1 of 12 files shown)


timothyb0912:
@bouzaghrane the function you want for simulating long format choices from a pylogit model is here: https://github.com/hassanobeid1994/tr_b_causal_2020/blob/8dc708340f4438c8a1c248d99ad0becf6bee9456/src/viz/utils.py#L225

An example of it in use after model estimation is here (see cell 7): https://github.com/adb40/machine_learning_session/blob/TimB-files/src/scripts/_03_tb-MNL_Checking.ipynb

timothyb0912:
@bouzaghrane I merged the pull request so you can import the functions directly now.

Issue #57, opened 2020-03-18
Selection-on-observables Simulation Goals Revisited


timothyb0912:
Hey @bouzaghrane and @hassanobeid1994 

I'm going to just jump straight to the topic at hand.

# The issue
As I mentioned via text message, I think that my original [idea](https://github.com/hassanobeid1994/tr_b_causal_2020/issues/5#issuecomment-584940240)—the idea driving the work that @bouzaghrane is doing—was wrong.

To recap, I thought that by manipulating the causal graph generating one’s explanatory variables, we could bias our parameter estimates in one’s outcome model.

Hassan raised the assertion that, no, as long as one’s outcome model is correctly specified, its parameters will be estimated unbiasedly, regardless of the generating model for X.

Though I originally disagreed, I now think Hassan was completely right and that my thought above was wrong.

# My rationale
Here is why I think Hassan is right and my earlier views in this issue are wrong.

The entire concept of exogeneity as it relates to causal inference is built around being able to infer the outcome model without needing to make inferences on the assignment model.

As explained by [Zhang, Zhang, and Scholkopf (2015)](https://arxiv.org/pdf/1504.05651.pdf), “the concept of exogeneity formalizes the idea that the mechanism generating the exogenous variable X does not contain any relevant information about the parameter set, phi, for the conditional model P(Y |X)”.

In other words, you can estimate the parameters of the outcome model without using any information from the model that generates X.

As shown in the Figure 1a of Zhang, Zhang, and Scholkopf (2015), this is precisely the unconfounded scenario.

# My Analysis
I think we ended up in this situation (with me holding erroneous ideas and @bouzaghrane perhaps doing some unnecessary work) for two reasons.

Reason one is that I never formally / explicitly justified why I thought my earlier views were correct.

Doing so would have, with higher probability, exposed the flaws in my thinking. In particular, my original logic was as follows.

Let A = Manipulating one’s causal graph, B = Multicollinearity and C = “serious problems” during outcome model estimation of GLMs.

Then since I know A can cause B and B can cause C, I concluded A can cause C.

The problem with this logic is that I didn’t explicitly define what these “serious problems” were. For some reason, I thought the “serious problem” was the introduction of parameter bias during estimation.

This is incorrect.

Multicollinearity inflates parameter’s standard errors and posterior dispersions, leading to parameter inferences that are close to the prior. In other words, multicollinearity does not lead to parameter bias, only statistical inefficiency.

[...]

PR #67, opened 2020-03-26
Predictive checks of deconfounder assumptions

timothyb0912:
This pull request initializes and will store files for performing predictive checks of one's causal model with latent variables.

The idea is that such checks serve two purposes.
- when using samples from the prior predictive distribution, they allow one to assess whether the observed data is at all consistent with your model assumptions, before estimation of one's posterior distribution.
- when using samples from the posterior predictive distribution, they allow one to ensure that one's observed data is consistent with one's model assumptions.

How this fits into the project is that along with conditional independence tests and marginal independence tests, this can be a third way to falsify one's causal graph. This approach will be generally applicable to **any** causal graph with latent variables.

This issue is related to the notion of what should we do to demonstrate how to falsify one's causal graph, as mentioned in #57 .


timothyb0912:
@bouzaghrane and @hassanobeid1994 for visibility.

timothyb0912:
@hassanobeid1994 I'm blocked on extracting the parameters of the variational approximate posterior distribution from the `confounder` function in `Causal_Graph_Tim_Data`.

Specific ask 1: Can you please checkout this branch, open the `_03-tb-deconfounder-predictive-checks.ipynb` notebook, run your `confounder` function on my `drive_alone_df.values`, and report back in a comment on this pull request `[w_mean_inferred, w_stddv_inferred, z_mean_inferred, z_stddv_inferred]`?

My issues seem mostly due to incompatible versions of tensorflow on my computer vs yours.

After this week, when we've finished prototypes of all the material for the presentation, we'll really need to create a comprehensive `environment.yml` file for this project.

I don't have the dependencies to run your notebooks (even after installing tensorflow and tensorflow probability) and it ruins our ability to collaborate since I can't execute your code.

Currently, other users besides just me will run into these issues. We need to make it such that anyone can immediately spin up a reproducible environment for our project and run ANY of our notebooks or `.py` files.

timothyb0912:
@hassanobeid1994,
besides the ask above, can you review my code please?

In particular, I want to make sure that we're in agreement over how I'm sampling from the deconfounder's factor model.

See the cells 8 and 9 in sections "Specify the factor model that is to be checked" and "Generate prior predictive samples".

Let me know if anything looks incorrect to you or if you have questions about anything I'm doing!

hassan-obeid:
@timothyb0912 I totally agree with you regarding the dependencies and we should do that in the coming weeks -- it was a pain to adjust my tensorflow versions to be compatible with Blei's version, so we should save people the trouble. 

I pushed two csv files to this same branch, one for W (mean and std) and one for Z (mean and std). They're in the same notebook working directory. Let me know if this unblocks you or if you need anything else. 

I'm reviewing your code today and will let you know if I have any comments/questions!

timothyb0912:
@hassanobeid1994 thank you thank you. This looks like it should be exactly what I need!

Thanks for getting to it quickly as well!

timothyb0912:
@hassanobeid1994 and @bouzaghrane 
Check out the files in [4847125](https://github.com/hassanobeid1994/tr_b_causal_2020/pull/67/commits/4847125404ef2776c3c8dcc650d113cd5b9b05c3), especially the [notebook](https://github.com/hassanobeid1994/tr_b_causal_2020/blob/4847125404ef2776c3c8dcc650d113cd5b9b05c3/notebooks/working/_03-tb-deconfounder-predictive-checks.ipynb).

Can both of you take a close look at my work and leave comments on the `.py` or `.md` files with
- questions about what I've done,
- recommendations for doing things in some better fashion,
- highlights of what you think I may be doing incorrectly if anything?

Assuming everything I've done in the notebook is correct, the conclusions are that:
1. The prior being used in our implementation (i.e. the naive implementation that anyone might reasonably start with) of the deconfounder severely conflicts our observed data.
2. The posterior used in our implementation of the deconfounder is also in great conflict with our observed data.
3. The predictive, conditional independence tests all fail because in general, the simulated data from the prior and posterior look nothing like the observed data.
4. Similar to Hassan's finding, the predictive conditional independence tests are only passed when the simulated confounder values are very close to the true latent confounder values. It would be good to see if this finding still holds true when using a factor model that fits the data well.

bouzaghrane:
@timothyb0912 will do today!

notebooks/working/_03-tb-deconfounder-predictive-checks.md
```diff
@@ -0,0 +1,683 @@
...
+num_dimensions = len(drive_alone_variables)
+
+# Specify the prior distributions for the factor
+# model of the standardized drive alone dataframe
+w_dist_prior = norm(loc=0, scale=1)
+z_dist_prior = norm(loc=0, scale=1)
+
+sigma_prior = 0.1
+epsilon_dists_prior =\
+    [norm(loc=0, scale=sigma_prior) for i in range(num_dimensions)]
```
bouzaghrane: Any reason why list comprehension is needed here vs. `num_dimensions * norm(loc=0, scale=sigma_prior)`?
Maybe each element in the list resulting from list comprehension has a different memory address?

notebooks/working/_03-tb-deconfounder-predictive-checks.md
```diff
@@ -0,0 +1,683 @@
...
+
+x_standardized_samples_prior =\
+    (np.einsum('mnr,ndr->mdr', z_samples_prior, w_samples_prior) +
+     epsilon_samples_prior)
+
+# Get samples of X on the original scale of each variable
+x_samples_prior =\
+    (x_standardized_samples_prior *
+     drive_alone_stds[None, :, None] +
+     drive_alone_means[None, :, None])
```
bouzaghrane: CLEAN!

notebooks/working/_03-tb-deconfounder-predictive-checks.md
```diff
@@ -0,0 +1,683 @@
...
+# Get prior samples of X_standardized
+w_samples_prior =\
+    w_dist_prior.rvs((1, num_dimensions, NUM_PRIOR_SAMPLES))
+z_samples_prior =\
+    z_dist_prior.rvs((num_drive_alone_obs, 1, NUM_PRIOR_SAMPLES))
+
+epsilon_samples_prior =\
+    np.concatenate(
+        [dist.rvs((num_drive_alone_obs, NUM_PRIOR_SAMPLES))[:, None, :]
+         for dist in epsilon_dists_prior],
```
bouzaghrane: The only question I have is about list comprehension again. Similar to my comment from above.

notebooks/working/_03-tb-deconfounder-predictive-checks.md
```diff
@@ -0,0 +1,683 @@
...
+# Get samples of X on the original scale of each variable
+x_samples_prior =\
+    (x_standardized_samples_prior *
+     drive_alone_stds[None, :, None] +
+     drive_alone_means[None, :, None])
+```
+
+```python
+# Look at the dimensions of the prior predictive samples
+print(x_samples_prior.shape)
```
bouzaghrane, line 397: No comment here, just wanted to say this is maybe what the shape of my simulated data should be. Store everything in a 3 dimensional array? We can discuss later.

notebooks/working/_03-tb-deconfounder-predictive-checks.md
timothyb0912: To be honest, I have no idea what multiplying a scalar by a distribution would do.

One thing I think I could have equivalently done is to simply set `epsilon_dists_prior = norm(loc=0, scale=sigma_prior)` and then get the correct dimensions when I sampled from it by doing `epsilon_dists_prior.rvs(size=(num_drive_alone_obs, num_dimensions, NUM_PRIOR_SAMPLES))`.

The list comprehension was the first thing that came to my mind because I knew for a fact that it would work. I def think this list comprehension should be replaced if possible when refactoring.

notebooks/working/_03-tb-deconfounder-predictive-checks.md
timothyb0912: This is wild. I never knew Github let one comment on multiple lines!

notebooks/working/_03-tb-deconfounder-predictive-checks.md
timothyb0912: Yep good call--it's likely the case that the list comprehension isn't needed and will get removed during refactoring!

notebooks/working/_03-tb-deconfounder-predictive-checks.md
timothyb0912: For sure. It's a little harder for your case as the number of observations isn't constant, so numpy wouldn't let you store things in a 3d array without some hacks like setting some rows to be all null or something similar.

notebooks/working/_03-tb-deconfounder-predictive-checks.md
bouzaghrane: Sorry what I meant was `num_dimensions * [norm(loc=0,scale=sigma_prior)]`

bouzaghrane:
Everything looks good to me @timothyb0912 .

bouzaghrane (commented):
Everything looks good to me @timothyb0912

notebooks/working/_03-tb-deconfounder-predictive-checks.md
timothyb0912: Oh, got ya. That would probably have the same effect!

timothyb0912:
@bouzaghrane: just made the changes to remove the list comprehension. Thanks for the call out!

PR #68, opened 2020-03-26
Ho investigating deconfounder

hassan-obeid:
Hey folks, 

My investigation of the deconfounder still shows that controlling for it is not helpful. Specifically, I simulated data in a way such that the coefficient on one of the variables will flip signs if the confounder was omitted. Adding the estimated confounder (estimated using 3 variations of PCA) did not help flip the sign back. Not sure what to focus my time on next.


timothyb0912:
@hassanobeid1994 can you remove the files from this branch that you're not actively editing?

I believe they may be present because you performed `git checkout -b"ho_investigating_deconfounder"` while you were on some branch other than master.

To remove files other than the deconfounder investigation files, I would do something like
```
cd tr_b_causal_2020
git checkout master -- references/organization/_01-ProblemStatement.md
git checkout master -- references/organization/_01-ProblemStatement.ipynb
git checkout master -- references/organization/_06-Project-Plan-and-Workflow.ipynb
git checkout master -- references/organization/_06-Project-Plan-and-Workflow.md
git add references/organization/_01-ProblemStatement.*
git add references/organization/_06-Project-Plan-and-Workflow.*
git commit -m"Removing files not being actively edited."
git push
```
Here, I'm assuming you weren't editing the problem statement or project plan as part of this pull request.

notebooks/working/investigating_deconfounder.py
```diff
@@ -0,0 +1,592 @@
...
+
+df = pd.DataFrame(np.array([a,b,c,d,e,f,confounder, y]).T, columns = ['a', 'b', 'c',
+          'd',  'e', 'f', 'confounder', 'y'])
+df.head()
+
+# +
+# sns.pairplot(df, size=1.5)
+
+# +
+ols_formula = "y ~ a+b+c+d+e+f"#+confounder "
```
timothyb0912: `model_full` should be full outcome model for `y` right?
```suggestion
ols_formula = "y ~ a+b+c+d+e+f+confounder "
```

notebooks/working/investigating_deconfounder.py
```diff
@@ -0,0 +1,592 @@
...
+        ("confounder", "b"),
+#         ("confounder", "c"),
+        ("confounder", "d"),
+        ("confounder", "e"),
+        ("confounder", "f"),
+
+
+        
+        
+        ("confounder", "y"),
```
timothyb0912: You can probably remove the repeated inclusion of `("confounder", "y")`.

notebooks/working/investigating_deconfounder.py
```diff
@@ -0,0 +1,592 @@
...
+results_df_partial['std_errors'] = results_partial.bse
+results_df_partial['true_params'] = coeffs_true[:-1]
+results_df_partial
+
+# ## Factor analysis: Is there evidence of latent factors?
+
+from factor_analyzer.factor_analyzer import calculate_bartlett_sphericity,calculate_kmo
+
+chi_square_value,p_value=calculate_bartlett_sphericity(df[['a','b','c','d','e','f']])
+print("Bartlett sphericity test indicates potential latent confounder, with a p-value of: ", p_value)
```
timothyb0912, line 181: According to wikipedia, the Bartlett test is testing for the assumption of equal variances amongst the k-column vectors. If one standardizes the variables inside the function before testing, then you can say the Bartlett test is testing for the assumption of k-independent, normally distributed vectors.

We should probably be specific since departures from normality may cause this test to fail, even if one has no unobserved latent confounders.

notebooks/working/investigating_deconfounder.py
```diff
@@ -0,0 +1,592 @@
...
+# Notice I designed the simulation such that ommitting the confounder will flip the sign of the coefficient on $b$. When we add the confounder, estimated using any PCA variation, the coefficient on $b$ remains flipped. 
+
+# +
+df['confounder_PCA_SKLEARN'] = principalComponents[:,0]
+
+df['confounder_PCA_CV'] = U[:,0]
+
+df['confounder_PPCA'] = confounders[2]
+# -
+
```
timothyb0912, line 362: Here, can you produce some visual checks of the inferred confounders?

I see from your results below that including the inferred confounders did not greatly change the regression results.

What I am now wondering is how good the inferred confounders are.

For each of your deconfounder methods (PCA, PPCA, PCA_CV), can you produce a scatter plot with actual confounder values on the x-axis, inferred confounder values on the y-axis, and a dashed diagonal line to indicate y = x?

Additionally, for the PPCA plot, if it's not too difficult, can you use the posterior predictive distribution of the confounder to show 95% credible interval bars on the inferred confounder values?

notebooks/working/investigating_deconfounder.py
```diff
@@ -0,0 +1,592 @@
...
+confounder = np.random.normal(loc=20, scale = 5, size = sample_size)
+
+# +
+a = .3*confounder + np.random.normal(loc=10,scale=3,size=sample_size)
+b = -1*confounder + np.random.normal(loc=6,scale=3,size=sample_size)
+c = np.random.normal(loc=15,scale=3,size=sample_size) # + .5*confounder 
+d = .8*confounder + np.random.normal(loc=-10,scale=3,size=sample_size)
+e = .5*confounder + np.random.normal(loc=8,scale=3,size=sample_size)
+f= -.3*confounder + np.random.normal(loc=-12,scale=3,size=sample_size)
+y = 5 + 2*a + 2*b - 5*c + 3*d + 2*e - 5*f + 7*confounder + np.random.normal(loc=0,scale=1,size=sample_size)
```
timothyb0912: So far, in the simulation shown in this notebook, you
- simulate 1 value of the confounder
- simulate 1 corresponding choice observation
- try to recover the confounder from that 1 choice observation

Based on plots of inferred vs actual confounders, my guess would be we are not at all recovering the confounders. Basically, we don't have enough information to identify (i.e. infer with any reasonable accuracy) the confounders.

Could you try re-running the simulation where you simulate multiple choice observations per individual with the same value of the latent confounder across all choice situations?

I imagine that as the number of observations per individual increases, our ability to recover the confounder for each individual will increase, and then the deconfounded estimates in the outcome model will more closely match the data generating parameters.

This would be similar to mixed logit models for panel data where our ability to infer latent, individual-specific parameters increases as we get more observations of each individual.

notebooks/working/investigating_deconfounder.py
```diff
@@ -0,0 +1,592 @@
...
+#       extension: .py
+#       format_name: light
+#       format_version: '1.5'
+#       jupytext_version: 1.3.4
+#   kernelspec:
+#     display_name: Python 3
+#     language: python
+#     name: python3
+# ---
+
```
timothyb0912, line 15: Can you list out the hypotheses / cases that you are planning on investigating in this notebook to generate evidence and information about when and why the deconfounder can be expected to work well?

It'll help readers understand what they're looking at and looking for in the notebook.

timothyb0912 (changes requested):
I did some more thinking about your finding that the confounder has to be recovered almost perfectly in order to generate outcome model parameter estimates that are close to the data-generating parameters.

I think a large part of the estimate volatility is that we're using a scalar summary statistic of the entire distribution of the deconfounder. Similar to mixed logit, if I estimated the model with only 1 draw (aka the mean) of the distribution of some latent variable, I would expect my estimate quality to be poor. If I expected otherwise, then I could just use the mean of my mixing distribution right?

Note that Wang and Blei **ONLY** justify using the point estimate if the posterior distribution concentrates around the point estimate. (See attached screenshot from their paper, second paragraph of section 4.2 from the journal paper or section 2.4.2 of the [arXiv version](https://arxiv.org/pdf/1805.06826.pdf)). The posterior distribution of the deconfounder (per individual) only concentrates at a point as the number of observations for each individual increases to infinity. So if we have 1 observation per individual, using a point estimate of the deconfounder is likely to be a bad idea.
Screen Shot 2020-03-28 at 7 02 36 PM


I propose that you test the hypothesis that (1) the lack of posterior concentration around the true confounder values and (2) the use of a point estimate of the deconfounder instead of using the posterior distribution are causing the outcome model estimation volatility.

1. Can you visualize a few (maybe 5 - 10 random) observations and show the PPCA posterior distributions of the deconfounder alongside the true values of the confounder?
2. Can you try estimating the outcome model by writing some custom code to perform the integration over the posterior distribution of the PPCA deconfounder instead of using just the mean of that distribution? Feel free to put this in a separate notebook if necessary to prevent this notebook from becoming too long and cluttered.

cc @bouzaghrane for visibility.

timothyb0912 (approved):

PR #77, opened 2020-04-02
adding more explanation and context to deconfounder investigation

hassan-obeid:
Tim -- can you please review and leave comments, especially for the section of "Factor Analysis: Is there evidence of latent factors"?


notebooks/working/investigating_deconfounder.py
```diff
@@ -14,7 +14,7 @@
 # ---
 
 # # Goal 
-# The goal of this notebook is to investigate, using simulations, the effectiveness and usefuleness of the deconfounder approach to causal inference with latent confounding in recovering the true causal effects of variables. 
+# The goal of this notebook is to investigate, using simulations, the effectiveness and usefuleness of the deconfounder approach to causal inference with latent confounding in recovering the true causal effects of variables. This approach was proposed by Blei et al. in 2018 (incert reference here)
```
timothyb0912, line 17:
```suggestion
# The goal of this notebook is to investigate, using simulations, the effectiveness and usefulness of the deconfounder approach to causal inference with latent confounding in recovering the true causal effects of variables. This approach was proposed by Wang and Blei in 2018 (insert reference here)
```

notebooks/working/investigating_deconfounder.py
```diff
@@ -158,6 +164,8 @@
 # -
 
 # ## Bias from ommitting the confounder without adjustment
+#
+# Here, we show that if we ommit the confounder, and attempt to estimate a new model for y, the coefficients obtained on each of the variables a, b, d, e, and f are all biased and do not represent the causal effect associated with each of those variables. The coefficient on c remains unbiased since c does not depend on the confounder. The purpose of this is to demonstrate that you indeed need to control for the confounder, or account for it somehow, if one knows you have confounded variables and is looking to assign causal interpretations to those variables
```
timothyb0912, line 168: ommit --> omit

notebooks/working/investigating_deconfounder.py
```diff
@@ -166,14 +174,22 @@
 results_df_full['true_params'] = coeffs_true
 results_df_full
 
-model_partial = smf.ols("y ~ a+b+c+d+e+f ", data=df)
+model_partial = smf.ols("y ~ a+b+c+d+e+f ", data=df) ## Ommit the confounder from OLS
```
timothyb0912, line 177:
```suggestion
model_partial = smf.ols("y ~ a+b+c+d+e+f ", data=df) ## Omit the confounder from OLS
```

notebooks/working/investigating_deconfounder.py
```diff
@@ -166,14 +174,22 @@
...
 results_df_partial = pd.DataFrame(results_partial.params, columns = ['estimated_params'])
 results_df_partial['std_errors'] = results_partial.bse
 results_df_partial['true_params'] = coeffs_true[:-1]
 results_df_partial
 
 # ## Factor analysis: Is there evidence of latent factors?
+#
+# How do we go about checking for whether a confounder exists (in this example, we know that a confounder exist since we defined the data generating process, but in the real world, the data generating process is more often than now unknown to the analyst). 
+#
+# There are a multitude of tests and ways to go about answering this question. A simple, off-the-shelf check is the bartlett sphericity test, which essentially just compares the correlation matrix of our covariates to the identity matrix. This test is used to check whether our covariate space can be reduced to a lower-dimensional space of factors that explain the variation in our data.
```
timothyb0912, line 188: Exactly as you wrote, the bartlett sphericity test only tests for whether one's correlation matrix is different from the identity matrix.

While a latent confounder might lead to a non-identity correlation matrix, it need not always do so since not all dependencies are linear. For instance, the correlation can be zero even if two random variables are dependent on one another. See https://en.wikipedia.org/wiki/Distance_correlation for examples and references.

Moreover, you can have correlations between random variables even if there are no latent confounders.

As a result, I would not use this test to determine if we have latent confounding. If one fails the test, then we know there is correlation between the variables but not whether latent confounding is causing the correlation. If we pass the test, we still don't know that there is no latent confounding.

notebooks/working/investigating_deconfounder.py
```diff
@@ -166,14 +174,22 @@
...
 
 # ## Factor analysis: Is there evidence of latent factors?
+#
+# How do we go about checking for whether a confounder exists (in this example, we know that a confounder exist since we defined the data generating process, but in the real world, the data generating process is more often than now unknown to the analyst). 
+#
+# There are a multitude of tests and ways to go about answering this question. A simple, off-the-shelf check is the bartlett sphericity test, which essentially just compares the correlation matrix of our covariates to the identity matrix. This test is used to check whether our covariate space can be reduced to a lower-dimensional space of factors that explain the variation in our data. 
+#
+# ### Tim, this would be a good place to add your other tests. I'm thinking of the bullet points below:
+# - We can hypothesize a causal graph, and run conditional independence tests. 
+# - Prior predictive checks?
```
timothyb0912, line 192: Yep, this is exactly where they should go.

notebooks/working/investigating_deconfounder.py
```diff
@@ -187,68 +203,23 @@
 
 # ## Fitting a factor model using 3 PCA variations.
 #
+# We fit three variations of PCA. This is not necessary indeed, but we wanted to see the robustness of our final estimates and their sensitivities to different variations of fitting the factor models.
```
timothyb0912, line 206:
```suggestion
# We fit three variations of PCA. This is not necessary needed, but we wanted to see the robustness of our final estimates and their sensitivities to different variations of fitting the factor models. 
```

notebooks/working/investigating_deconfounder.py
```diff
@@ -338,28 +309,39 @@
...
 df['confounder_PPCA'] = confounders[2]
 # -
 
+df.plot(kind='scatter', x = 'confounder_PCA_SKLEARN', y = 'confounder')
+
+df.plot(kind='scatter', x = 'confounder_PPCA', y = 'confounder')
+
+# ## Re-estimating regression with and without confounder
+#
+# Notice that the simulation was designed such that ommitting the confounder will flip the sign of the coefficient on $b$. When we add the confounder, estimated using any PCA variation, the coefficient on $b$ remains flipped.
```
timothyb0912, line 341:
```suggestion
# Notice that the simulation was designed such that omitting the confounder will flip the sign of the coefficient on $b$. When we add the confounder, estimated using any PCA variation, the coefficient on $b$ remains flipped. 
```

notebooks/working/investigating_deconfounder.py
```diff
@@ -327,7 +298,7 @@
 display.display(sns.kdeplot(pvals[holdout_subjects]))
```
timothyb0912, line 298: Here, can you make the point horrendously obvious to viewers that the overall p-value is meaningless in this case.

Specifically, can you just plot the vertical line of the overall p-value on the plot with a label that says something along the lines of `"Observed: {:.2f}".format(pval_mode)`

notebooks/working/investigating_deconfounder.py
```diff
@@ -338,28 +309,39 @@
     ax.axvline(x=obs_ll_per_zi[subject_no], linestyle='--')
 
 plt.show()
```
timothyb0912, line 311: Underneath the plots, can you add some text about the fact that Wang and Blei only justify the use of the mean of the inferred deconfounder (i.e. `z_inferred_mean`) if the posterior of the substitute confounder concentrates at a point.

Feel free to copy their statement directly from their article or put a direct screenshot from their paper (probably even better). Given the images above, two things should be clear to readers:
1. that we are not justified in using the mean of the substitute confounder in our application
2. that this sort of visual analysis (alongside more quantitative analyses such as looking at the variance of `z_inferred` across observations) should be performed to assess convergence-to / concentration-around a point.

To make it even easier, feel free to take from my comments on your other pull request (#68 ) where I first brought this up.

notebooks/working/investigating_deconfounder.py
```diff
@@ -338,28 +309,39 @@
...
-# ## Visualizing the substitute confounder against the true confounder
+# +
+overall_ll_est = rep_ll_per_zi.mean(axis=1)
 
-df.plot(kind='scatter', x = 'confounder_PCA_SKLEARN', y = 'confounder')
+overall_ll_obs = obs_ll_per_zi.mean()
 
-df.plot(kind='scatter', x = 'confounder_PPCA', y = 'confounder')
+sns.kdeplot(overall_ll_est).set_title("Predictive check: Population")
+plt.axvline(x=overall_ll_obs, linestyle='--')
```
timothyb0912, line 319: Can you add some text under this plot to let readers know that this is yet another posterior predictive check of the factor model's log-likelihood, and the predictive check fails terribly?

Just point out that a model should not be strongly refuted by any posterior predictive checks, and if it is, it should cause one to further investigate one's model performance (even if another posterior predictive check has passed).

notebooks/working/investigating_deconfounder.py
```diff
@@ -187,68 +203,23 @@
 
 # ## Fitting a factor model using 3 PCA variations.
```
timothyb0912, line 204: Am I correct in seeing you only use 2 variants not 3?

notebooks/working/investigating_deconfounder.py
```diff
@@ -389,28 +371,36 @@
 results_df_deconf
 
 # ## Adding random noise to the confounder and seeing how it affects the coefficients on the other variables
```
timothyb0912, line 375: In this section, can you produce a plot where the x-axis is the standard deviation of the noise that is added to the confounder and the y-axis is the estimated coefficient value on the new variable (plus error bars coming from the confidence interval around the coefficient)? Also on the plot should be a horizontal line at the true coefficient value (7 in this case).

I think it will be helpful to show the estimation results breaking down with more noise to see roughly how much noise can be added before we stop getting reasonable estimates.

When describing how much noise can be added before the estimation breaks down, it may be helpful to express it as a percentage `(std_dev_noise / std_dev_on_confounder) * 100`.

timothyb0912 (changes requested):
Hey Hassan! I took a quick look!

You can see the individual comments above, but the high-level / important things I would change would be:
- add descriptions under the plots of the posterior predictive checks and inferred confounder distributions
- add plots related to seeing when the confounder breaks down.

Otherwise this mostly looks good. I'll come back and give some final comments when I'm not running to dinner! =p.

timothyb0912 (changes requested):
Also, I wonder if it would help to use a deliberately misspecified model (but one that has higher predictive power)?

In the deconfounder paper, they use the predicted vector of reconstructed causes in one example. This leads me to think that you may be able to get good causal inferences if the reconstructed causes are modeled well, even if the substitute confounder that you get isn't accurate (since the model is misspecified).
Screen Shot 2020-04-03 at 7 57 20 AM

Could you try re-running your regressions using a factor model with 2 and 3 latent dimensions?
Note there is precedent for misspecified models working very well (especially overparameterized ones, and especially in the context of latent variable models).

I can provide citations later (or google scholar can be your friend immediately =p).

timothyb0912:
Some citations include:
- Gelman, Andrew, et al. "[Using redundant parameterizations to fit hierarchical models](https://wwwf.imperial.ac.uk/~dvandyk/Research/08-jcgs-pxda.pdf)." Journal of Computational and Graphical Statistics 17.1 (2008): 95-122.
- Arora, Sanjeev, Nadav Cohen, and Elad Hazan. "[On the optimization of deep networks](https://arxiv.org/pdf/1802.06509.pdf): Implicit acceleration by overparameterization." arXiv preprint arXiv:1802.06509 (2018).
- Xu, J., Hsu, D. J., & Maleki, A. (2018). [Benefits of over-parameterization with EM](http://papers.nips.cc/paper/8265-benefits-of-over-parameterization-with-em.pdf). In Advances in Neural Information Processing Systems (pp. 10662-10672). 
- Buhai, R. D., Risteski, A., Halpern, Y., & Sontag, D. (2019). [Benefits of overparameterization in single-layer latent variable generative models](https://arxiv.org/pdf/1907.00030.pdf). arXiv preprint arXiv:1907.00030.
- https://www.yulingyao.com/blog/2019/stacking/

PR #87, opened 2020-04-09
Ho public notebook skeleton

hassan-obeid:
I added a first draft of the skeleton of the external notebook. It only covers our work for step 1 so far. I'm curious about your thoughts on whether the work on the deconfounder should be part of the same notebook, or should we create another notebook for that.


notebooks/working/external_notebook_skeleton.ipynb
```diff
@@ -0,0 +1,219 @@
...
+    "- We perturb travel distance here, and only change travel distance\n",
+    "- Apply the predict function of our outcome model. \n",
+    "- Store the change in the mode share of auto. "
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### True approach\n",
```
timothyb0912, line 145: nit
```suggestion
    "### Correct approach\n",
```

notebooks/working/external_notebook_skeleton.ipynb
```diff
@@ -0,0 +1,219 @@
...
+   "source": []
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Re-run the approach above N times and plot visual comparisons\n",
+    "The sections above make the point we're trying to make. Here I envision adding some text saying that we re-run the above process N times, and store the changes.\n",
+    "\n",
+    "I imagine having a cell with code that simply puts the code from the previous section in a loop and just repeats it N times. \n",
```
timothyb0912, line 185: For the public notebook, I wouldn't even show the code that repeats the causal effect estimation N times. As seen in Amine's notebook, that code is a lot to read. I would instead suggest simply calling a function that is named clearly...something like `repeat_causal_effect_estimation(NUM_SIMULATIONS)`

notebooks/working/external_notebook_skeleton.ipynb
```diff
@@ -0,0 +1,219 @@
...
+    "\n",
+    "The second cell plots the distributions of each of the effects calculated using each of the three approaches. "
```
timothyb0912, line 187: I think we're together on this cell, but just to specify, I'd suggest using a helper function that is clearly named, e.g. `plot_comparison_of_effect_estimators(naive_effects, correct_effects, true_effects)`.

notebooks/working/external_notebook_skeleton.ipynb
```diff
@@ -0,0 +1,219 @@
...
+    "- Apply the predict function of *the original model used for simulation* on the new set of covariates\n",
+    "- Store the change in the mode share of auto"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Make conclusion\n",
+    "Here we just compare the magnitude of the three causal effects obtained above, using some t-test or simply visually"
```
timothyb0912, line 168: At this point, we'll have only calculated the causal effects once using each method (naive, correct, and true). I think it'll be hard to do any t-test as we won't have standard error estimates. My guess is that simply showing the three estimates visually will be enough, especially since we'll show the distributions two cells down.

notebooks/working/external_notebook_skeleton.ipynb
```diff
@@ -0,0 +1,219 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": []
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Simulating the data\n",
+    "Brief paragraph description of this section, e.g. outline that we'll simulate the Xs in two ways etc. "
```
timothyb0912, line 80: Am I correct in thinking that we agreed to not simulate the data in two ways for the public notebook? I think we agreed to only simulate the data from the true causal graph and to calculate the causal effects in two ways:
- by only perturbing travel distance
- by perturbing travel distance and its downstream nodes.

timothyb0912: Here, I think we should just describe what's going on internally below: estimating the distributions of each node conditioned on its parents, based on our causal graph, using the real data.

notebooks/working/external_notebook_skeleton.ipynb
```diff
@@ -0,0 +1,219 @@
...
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": []
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Description of the dataset -- written or using some descriptive stats"
```
timothyb0912, line 51: I would also suggest that we show a couple of rows of the dataframe. That always helps me understand what the data looks like. Imagine `df.head(5).T`. The transpose is in case the dataframe has many columns (which I believe ours does). Also, can we only show the rows being used in the analysis? I don't think it should matter much, but to be sure, can we not include extraneous rows (e.g. the various household id columns and such)?

notebooks/working/external_notebook_skeleton.ipynb
```diff
@@ -0,0 +1,219 @@
...
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": []
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Set work directory and load the dataset. "
```
timothyb0912, line 37: Just for good practice, can you designate an earlier cell to store notebook parameters / global constants? Things like filepaths to data, the number of simulations to be executed, random seeds, filepaths to store the plots created at the end, etc.

timothyb0912 (changes requested):
1. Can you pair a markdown or py file with the `external_notebook_skeleton` for easy github review / comment?
2. I assume this branch is only meant to create the `external_notebook_skeleton` right? If so, could you remove the `investigating_deconfounder` files from this branch? I believe this could be done with `git rm investigating_deconfounder.*` and `git rm references/organization/Project_story_outline.*`. Just make sure they're indeed in some other branch!
3. I assume the external_notebook_skeleton is only for the selection on observables notebook right? In this case, it should likely be renamed something like `selection_on_observables_public_skeleton`.
   1. I assume there'll be similar skeletons for the other public notebooks? (See the public notebooks section here, https://github.com/hassanobeid1994/tr_b_causal_2020/blob/master/references/organization/_05-Requirements.md).
   2. Feel free to not make a skeleton for the notebook demonstrating the falsification techniques if it makes life easier, and then you can just review the notebook after I make it. Or you can make one and I'll give some comments and follow your final outline. Whichever works best for you!

bouzaghrane:
Once you make the suggested edits @hassan-obeid, can you move the notebook to a final stage so I can try to run all notebooks and see if I can reproduce everything?

PR #89, opened 2020-04-11
MB Refactoring

bouzaghrane:
This branch is created from the causal-effect-estimation branch.

@timothyb0912 @hassanobeid1994


bouzaghrane:
Thinking about the suggestions I have, I see that they have to purposes:
* Make each function have one simple functionality
* Minimize duplication across functions and code in general

bouzaghrane:
@timothyb0912 I just completely refactored the DistNodeNoParent function and wrote a test for it at the bottom of the notebook. The test is based on `unittest` and testing modules from `numpy`. Once I put all of my functions in a `.py` file, I will make the transition from `unittest` to `pytest` which should be very straightforward.

notebooks/working/3.0-mab-independent-variable-simulation.md
```diff
@@ -34,487 +47,269 @@ import random
...
+    cont_dict[ind_var]['distribution'] = list(BestDict.items())[0][0]
+    cont_dict[ind_var]['parameters'] = list(BestDict.items())[0][1]
+    return cont_dict
+
+## Function for looping around individual specific variables
+def indSpecDist(data_long, obs_id_col, ind_spec, var_types):
+    ind_spec_dict = defaultdict(dict)
+    for ind_var in ind_spec:
+        # generate array of values for individual specific variable
+        var_val = pd.Series([(data_long.loc[data_long[obs_id_col] == x][ind_var].unique()[0]) for x in data_long[obs_id_col].unique()])
```
timothyb0912: Can you shorten the length of this line?
```suggestion

        def extract_indiv_specific_data(obs_id, ind_var_col):
            indiv_mask = data_long[obs_id_col] == obs_id
            indiv_data_value =\
                data_long.loc[indiv_mask, ind_var_col].iloc[0]
            return indiv_data_value

        var_val =\
            pd.Series([extract_indiv_specific_data(x, ind_var)
                       for x in data_long[obs_id_col].unique()])
```

And note, I'd put the definition of `extract_indiv_specific_data` outside the for-loop.

timothyb0912: Actually, couldn't you (perhaps even more simply), generate `var_val` with some pandas manipulation as follows:
```suggestion
        var_val =\
            (data_long[[obs_id_col, ind_var]]
                .drop_duplicates(obs_id_col, inplace=False)
                .loc[:, ind_var]
                .reset_index(drop=True)
            )
```

notebooks/working/3.0-mab-independent-variable-simulation.md
```diff
@@ -34,487 +47,269 @@ import random
...
-```
+# Code for Trip Specific Variables
+# Loop around the different available alternatives
+def tripSpecDist(data_long, obs_id_col, trip_spec, cont_dists):
+    # Trip Specific Variable (maybe combine with individual specific variables)
+    # Loop around trip (observation) specific variables
+    trip_spec_dict = defaultdict(dict)
+    for trip_var in trip_spec:
+        # generate array of values for trip specific variable
+        trip_var_data = pd.Series([(data_long.loc[data_long[obs_id_col] == x][trip_var].unique()[0]) for x in data_long[obs_id_col].unique()])
```
timothyb0912: Similar comment to as above. Can you shorten the line length to 79 characters per line (max)?

Also here's a suggestion for generating trip_var_data.
```suggestion
        trip_var_data =\
            (data_long[[obs_id_col, trip_var]]
                .drop_duplicates(obs_id_col, inplace=False)
                .loc[:, trip_var]
                .reset_index(drop=True)
            )
```

notebooks/working/3.0-mab-independent-variable-simulation.md
```diff
@@ -34,487 +47,269 @@ import random
...
+# this functionality will be relocated to
+# a different function, most likely a function
+# that merges availability matrix and simulated
+# data or a function that converts to long_data
+def createFakeChoiceCol(AV_matrix):
+    # Create an random choice column based on available
+    # alternatives for each observation - This column will
+    # be needed when converting to long data -- this will
+    # be moved to a different column
+    fake_choice = [random.choice(np.nonzero(a == 1)[0]) + 1 for a in np.array(AV_matrix)]
```
timothyb0912: I find this line hard to understand upon first reading it.

I would separate the various parts of this line to show what one desires with each action. For example,
```suggestion
    available_alt_indices = lambda av_row: np.nonzero(av_row == 1)[0]
    fake_choice = [random.choice(available_alt_indices(a)) + 1 for a in np.array(AV_matrix)]
```

notebooks/working/3.0-mab-independent-variable-simulation.md
```diff
@@ -34,487 +47,269 @@ import random
...
+
+## Functions to replace code for getting
+## distributions for alternative specific
+## variables
+
+def getConstantDistAlt(alt_var, alt_data, alt_name_dic, alt):
+    constant_dict = defaultdict(dict)
+    # Add name of alternative to variable and store distriburion & parameters
+    constant_dict[alt_var+'_'+alt_name_dic[alt]]['distribution'] = 'constant'
+    constant_dict[alt_var+'_'+alt_name_dic[alt]]['parameters'] = alt_data[alt_var].unique()
```
timothyb0912: To the best extent possible, try to not repeat yourself. For example,
```suggestion
    alt_specific_var_name = alt_var + '_' + alt_name_dic[alt]
    constant_dict[alt_specific_var_name]['distribution'] = 'constant'
    constant_dict[alt_specific_var_name]['parameters'] = alt_data[alt_var].unique()
```

timothyb0912: As a matter of fact, you end up repeating this same pattern across multiple functions (for example, in `getCategoricalDistAlt` and `getContinuousDistAlt`).

This suggests that you should separate this logic into its own function. One might do:
```suggestion
    alt_specific_var_name =\
        get_alt_specific_variable_name(alt_var, alt_name_dict, alt)
    constant_dict[alt_specific_var_name]['distribution'] = 'constant'
    constant_dict[alt_specific_var_name]['parameters'] = alt_data[alt_var].unique()
```
You can define the separated function as follows:
```
def get_alt_specific_variable_name(var_name, alt_name_map, alt_id):
    return var_name + '_' + alt_name_map[alt_id]
```

notebooks/working/3.0-mab-independent-variable-simulation.md
```diff
@@ -34,487 +47,269 @@ import random
...
+    return constant_dict
+
+def getCategoricalDistAlt(alt_var, alt_data, alt_name_dic, alt):
+    categ_dict = defaultdict(dict)
+    categ_dict[alt_var+'_'+alt_name_dic[alt]]['distribution'] = 'categorical'
+    np_array_range = np.arange(alt_data[alt_var].max()+1)
+    array_bincount = np.bincount(alt_data[alt_var])
+    probs = array_bincount / len(alt_data[alt_var])
+    categ_dict[alt_var+'_'+alt_name_dic[alt]]['parameters'] = [np_array_range,
+                                                               probs]
```
timothyb0912:
```suggestion
    alt_specific_var_name =\
        get_alt_specific_variable_name(alt_var, alt_name_dict, alt)
    categ_dict[alt_specific_var_name]['parameters'] =\
        [np_array_range, probs]
```

notebooks/working/3.0-mab-independent-variable-simulation.md
```diff
@@ -34,487 +47,269 @@ import random
...
+    # to the data
+    fitter_object = Fitter(data=alt_data[alt_var],
+                           distributions=cont_dists,
+                           timeout=30)
+    fitter_object.fit()
+    # Get the best distribution and store in dictionary
+    BestDict = fitter_object.get_best()
+    # Add name of alternative to variable and store distriburion & parameters
+    cont_dict[alt_var+'_'+alt_name_dic[alt]]['distribution'] = list(BestDict.items())[0][0]
+    cont_dict[alt_var+'_'+alt_name_dic[alt]]['parameters'] = list(BestDict.items())[0][1]
```
timothyb0912: Same comment as above re the logic for creating the alternative specific variable name.

notebooks/working/3.0-mab-independent-variable-simulation.md
```diff
@@ -34,487 +47,269 @@ import random
...
+## Function for checking length
+def isUnique(var_values):
+    return len(var_values.unique()) == 1
+
+## Function for checking if variable is categorical
+def isCategorical(var_type):
+    return var_type == 'categorical'
+
+## Function to get distribution for unique values:
+def getConstantDist(ind_var, var_val):
```
timothyb0912: I would increase the generality of this function as well as `getCategoricalDist` and `getContinuousDist`.
```suggestion
def getConstantDist(var_name, var_val):
```

notebooks/working/3.0-mab-independent-variable-simulation.md
```diff
@@ -34,487 +47,269 @@ import random
...
+## variables
+
+def getConstantDistAlt(alt_var, alt_data, alt_name_dic, alt):
+    constant_dict = defaultdict(dict)
+    # Add name of alternative to variable and store distriburion & parameters
+    constant_dict[alt_var+'_'+alt_name_dic[alt]]['distribution'] = 'constant'
+    constant_dict[alt_var+'_'+alt_name_dic[alt]]['parameters'] = alt_data[alt_var].unique()
+    return constant_dict
+
+def getCategoricalDistAlt(alt_var, alt_data, alt_name_dic, alt):
```
timothyb0912: I would get rid of the following functions:
1. `getConstantDistAlt`
2. `getCategoricalDistAlt`
3. `getContinuousDistAlt`

They can be combined with their corresponding functions without the `Alt` if you abstract the generation of the alternative specific variable name into a separate function as suggested [above](https://github.com/hassanobeid1994/tr_b_causal_2020/pull/89/files?file-filters%5B%5D=.md#r419154494).

notebooks/working/3.0-mab-independent-variable-simulation.md
```diff
@@ -34,487 +47,269 @@ import random
...
+# Loop around the different available alternatives
+def tripSpecDist(data_long, obs_id_col, trip_spec, cont_dists):
+    # Trip Specific Variable (maybe combine with individual specific variables)
+    # Loop around trip (observation) specific variables
+    trip_spec_dict = defaultdict(dict)
+    for trip_var in trip_spec:
+        # generate array of values for trip specific variable
+        trip_var_data = pd.Series([(data_long.loc[data_long[obs_id_col] == x][trip_var].unique()[0]) for x in data_long[obs_id_col].unique()])
+        # Get distribution if variable is categorical
+        var_type = var_types[trip_var]
```
timothyb0912, line 269: I believe you are relying on a global variable here (which you shouldn't do).
`var_types` is not defined as one of the function inputs, and it is not generated between lines 199 and 205. Try adding `var_types` to your inputs!

notebooks/working/3.0-mab-independent-variable-simulation.md
```diff
@@ -34,487 +47,269 @@ import random
...
+        # Get distribution if variable is categorical
+        if isCategorical(var_type):
+            # If only one category
+            if isUnique(var_val):
+                ind_var_dict = getConstantDist(ind_var, var_val)
+            # If more than one category
+            else:
+                ind_var_dict = getCategoricalDist(ind_var, var_val)
+        else:
+            ind_var_dict = getContinuousDist(ind_var, var_val, cont_dists)
```
timothyb0912: I believe you are relying on a global variable here (which you shouldn't do).
`cont_dists` is not defined as one of the function inputs, and it is not generated between lines 100 and 114. Try adding `cont_dists` to your inputs!

notebooks/working/3.0-mab-independent-variable-simulation.md
```diff
@@ -34,487 +47,269 @@ import random
...
-#                          4: 'wtw',
-#                          5: 'dtw',
-#                          6: 'wtd',
-#                          7: 'walk',
-#                          8: 'bike'}
-```
+# Code for Trip Specific Variables
+# Loop around the different available alternatives
+def tripSpecDist(data_long, obs_id_col, trip_spec, cont_dists):
+    # Trip Specific Variable (maybe combine with individual specific variables)
```
timothyb0912, line 257: 👍 Yes please combine the two functions.

If you truly did have another identifier for trip-id's, then (I assume?) they would follow long-format conventions of not repeating identifying information across columns. In other words, simply passing in a `trip_id_col` would not be sufficient to filter the dataframe to 1 row per decision-maker per trip. You would need the `obs_id_col` and the `trip_id_col`, and you would need to change the logic used to get the data that you want.

As such, this function doesn't do what you would want it to do (get a distribution of values across all unique trips). At the moment, it only duplicates the logic that should be used in `indSpecDist`.

notebooks/working/3.0-mab-independent-variable-simulation.md
```diff
@@ -34,487 +47,269 @@ import random
...
+            else:
+            # If more than one category
+                trip_spec_var_dic = getCategoricalDist(trip_var, trip_var_data)
+        else:
+            # If not categorical but just one unique value
+            if isUnique(trip_var_data):
+                trip_spec_var_dic = getConstantDist(trip_var, trip_var_data)
+            # If not categorical but just one unique value
+            else:
+                trip_spec_var_dic = getContinuousDist(trip_var, trip_var_data, cont_dists)
```
timothyb0912: This logic exists in `indSpecDist`, `altSpecDist`, and `tripSpecDist`.

It should be separated into its own function `get_distribution_and_param_dict(var_name, var_data, cont_dists=ALL_SCIPY_CONT_DISTS)`, with `ALL_SCIPY_CONT_DISTS` defined as a global variable.

timothyb0912 (changes requested):
Hey @bouzaghrane !
Here are my (initial) overall comments on Notebook 3 in your PR.

## (General) Suggestions
1. Follow PEP8 style guidelines unless there's a good reason not too.
   1. Use snake casing rather than camel-casing for your function names.
   2. Adhere to line-lengths of 79 characters per line (so people can easily read your code when splitting screens).
2. See whether you can combine `altSpecDist` and `indSpecDist` into a single function `get_distribution_dicts`. Skip the suggestion if it's harder than is worth the effort, but my strong feeling is that these two functions are more similar than they are different. In particular, you may be able to abstract to a single function, with a kwarg of `name_dict=None`. Then within the secondary for-loop of the general function, you can have something along the lines of `var_name = spec_key if name_dict is None else name_dict[spec_key]`.
3. After the final set of functions is nailed down, add docstrings to all of them.

notebooks/working/6.0-mab-estimate-causal-effect.md
```diff
@@ -40,443 +66,628 @@ import viz
...
+            # If not categorical but just one unique value
+            else:
+                trip_spec_var_dic = getConstantDist(trip_var, trip_var_data, cont_dists)
+        trip_spec_dict.update(trip_spec_var_dic)
+    return trip_spec_dict
 
-# 1. Define functions to be used in notebook
 
-```python
+## Define the main function
```
timothyb0912: The entire `DistNodeNoParent` function is way cleaner than before. Good job! 👍

notebooks/working/6.0-mab-estimate-causal-effect.md
```diff
@@ -536,151 +747,71 @@ def DistNodeNoParent(data_long,
...
-                                       distributions=cont_dists,
-                                       timeout=60)
-                fitter_object.fit()
-                # Get the best distribution and store in dictionary
-                BestDict = fitter_object.get_best()
-                params_dict.setdefault(var, {})['distribution'] = list(BestDict.items())[0][0]
-                params_dict.setdefault(var, {})['parameters'] = list(BestDict.items())[0][1]
-
+    trip_spec_dic_params = tripSpecDist(data_long, obs_id_col, trip_spec, var_types, cont_dists)
+    params_dict.update(trip_spec_dic_params)
```
timothyb0912: Similar to my comment from notebook 3, can you combine the various functions into a new function,
`get_distribution_dicts`, with a kwarg `name_dict=None`? Then within the second for-loop of one's function, you can have a line that says `var_name = spec_key if name_dict is None else name_dict[spec_key]`. I believe that's the main differences between `indSpecDist`, `altSpecDist`, and `tripSpecDist`.

notebooks/working/6.0-mab-estimate-causal-effect.md
```diff
@@ -40,443 +66,628 @@ import viz
...
+            # If more than one category
+            else:
+                trip_spec_var_dic = getCategoricalDist(trip_var, trip_var_data)
+        else:
+            # If not categorical but just one unique value
+            if isUnique(trip_var_data):
+                trip_spec_var_dic = getConstantDist(trip_var, trip_var_data)
+            # If not categorical but just one unique value
+            else:
+                trip_spec_var_dic = getConstantDist(trip_var, trip_var_data, cont_dists)
```
timothyb0912: Across all three functions, `indSpecDis`, `altSpecDist`, and `tripSpecDist`, the logic in the highlighted lines is the same. This logic should be refactored out into a separate function, perhaps `get_distribution_and_param_dict(var_name, var_values, cont_dists=ALL_SCIPY_CONT_DISTS)` where you define `ALL_SCIPY_CONT_DISTS` as a global constant.

notebooks/working/6.0-mab-estimate-causal-effect.md
```diff
@@ -40,443 +66,628 @@ import viz
...
+def tripSpecDist(data_long, obs_id_col, trip_spec, var_types, cont_dists):
+    """
+    Function that retrieves distributions for all trip
+    specific variables.
+    """    # Trip Specific Variable (maybe combine with individual specific variables)
+    # Loop around trip (observation) specific variables
+    trip_spec_dict = {}
+    for trip_var in trip_spec:
+        # generate array of values for trip specific variable
+        trip_var_data = pd.Series([(data_long.loc[data_long[obs_id_col] == x][trip_var].unique()[0]) for x in data_long[obs_id_col].unique()])
```
timothyb0912: Same suggestion as notebook 3:
```suggestion
        trip_var_data =\
            (data_long[[obs_id_col, trip_var]]
                .drop_duplicates(obs_id_col, inplace=False)
                .loc[:, trip_var]
                .reset_index(drop=True)
            )
```

notebooks/working/6.0-mab-estimate-causal-effect.md
```diff
@@ -40,443 +66,628 @@ import viz
...
+def isEmpirical(var_type):
+    """
+    Checks whether the variable type for the
+    variable of interest is to be taken
+    as a constant value or as numerical values.
+    """
+    return var_type == 'empirical'
 
-#     mapping_matrix = sparse_row_to_col_matrix.dot(sparse_row_to_col_matrix.T)
+def getConstantDistAlt(alt_var, alt_data, alt_name_dic, alt):
```
timothyb0912: As in notebook 3, I'd suggest getting rid of all functions `getConstantDistAlt`, `getEmpiricalDistAlt`, `getCategoricalDistAlt`, and `getContinuousDistAlt` in favor of their corresponding functions without the `Alt` suffix.

To do this, can you make all the corresponding functions simply take `(var_name, var_data)` as inputs? Then, for the alternative specific cases, you can use the function
```
def get_alt_specific_variable_name(base_var_name, alt_name_map, alt_id):
    return base_var_name + '_' + alt_name_map[alt_id]
```
on the lines before you generate the dictionary of distribution name and parameters.

notebooks/working/6.0-mab-estimate-causal-effect.md
```diff
@@ -40,443 +66,628 @@ import viz
...
-# def calculate_utilities(betas, design_matrix):
+# 1. Define functions to be used in notebook
 
-#     limit_max = 700
-#     limit_min = -700
+```python
+## Functions to replace code within
+## DistNodeNoParent
+## Function for checking length
+def isUnique(var_values):
```
timothyb0912: Can you camel case your function names? Also, feel free to let me know if there's a reason to not do so that I'm missing!

notebooks/working/6.0-mab-estimate-causal-effect.md
```diff
@@ -40,443 +66,628 @@ import viz
...
+
+def indSpecDist(data_long, obs_id_col, ind_spec, var_types, cont_dists):
+    """
+    Function that retrieves distributions for all individual
+    specific variables.
+    """
+    ind_spec_dict = {}
+    for ind_var in ind_spec:
+        # generate array of values for individual specific variable
+        var_val = pd.Series([(data_long.loc[data_long[obs_id_col] == x][ind_var].unique()[0]) for x in data_long[obs_id_col].unique()])
```
timothyb0912: From notebook 3:
```suggestion
        var_val =\
            (data_long[[obs_id_col, ind_var]]
                .drop_duplicates(obs_id_col, inplace=False)
                .loc[:, ind_var]
                .reset_index(drop=True)
            )
```

notebooks/working/6.0-mab-estimate-causal-effect.md
```diff
@@ -704,43 +835,192 @@ def SimNodeNoParent(params_dict, size=1000):
...
+# this functionality will be relocated to
+# a different function, most likely a function
+# that merges availability matrix and simulated
+# data or a function that converts to long_data
+def createFakeChoiceCol(AV_matrix):
+    # Create an random choice column based on available
+    # alternatives for each observation - This column will
+    # be needed when converting to long data -- this will
+    # be moved to a different column
+    fake_choice = [random.choice(np.nonzero(a == 1)[0]) + 1 for a in np.array(AV_matrix)]
```
timothyb0912: To make this function easier to understand, I suggest the following.
It'll make it clear what `np.nonzero(a == 1)[0]` is doing.
```suggestion
    available_alt_indices = lambda av_row: np.nonzero(av_row == 1)[0]
    fake_choice = [random.choice(available_alt_indices(a)) + 1 for a in np.array(AV_matrix)]
```

timothyb0912 (changes requested):
This is a review of the portion of notebook 6 before the simulation.

The comments are mainly the same as my review of notebook 3.

Overall, I greatly agree with your general plan of what else needs to be refactored.
I would add an effort to at times combine functionality when possible. It's probably better to have a function that is trivially generalized by one or two kwargs, if it allows one to save duplicating large parts of two functions.

notebooks/working/3.0-mab-independent-variable-simulation.md
bouzaghrane: You're right! Much faster as well!

notebooks/working/3.0-mab-independent-variable-simulation.md
bouzaghrane: Thanks!

notebooks/working/3.0-mab-independent-variable-simulation.md
bouzaghrane: Very true.

notebooks/working/3.0-mab-independent-variable-simulation.md
bouzaghrane: Got it! Whenever a logic repeats, isolate it.

notebooks/working/3.0-mab-independent-variable-simulation.md
bouzaghrane: Yes sir!

notebooks/working/3.0-mab-independent-variable-simulation.md
bouzaghrane: 👍

notebooks/working/3.0-mab-independent-variable-simulation.md
bouzaghrane: Good catch! totally missed that.

notebooks/working/3.0-mab-independent-variable-simulation.md
bouzaghrane: Great catch!

notebooks/working/3.0-mab-independent-variable-simulation.md
bouzaghrane: Agreed. In this case, is our `tour_id` variable the `trip_id_col` we want? The confusion is just around whether the `tour_id` can be the same for multiple trips on the same tour or not.

notebooks/working/3.0-mab-independent-variable-simulation.md
bouzaghrane: Agreed, I just need to figure out how to combine all the kwargs from different functions given that we need different arguments in different functions.

notebooks/working/6.0-mab-estimate-causal-effect.md
bouzaghrane: Yes!

notebooks/working/6.0-mab-estimate-causal-effect.md
bouzaghrane: No specific reason. I'll snake case them.

notebooks/working/6.0-mab-estimate-causal-effect.md
bouzaghrane: Actually, I think that only the logic for finding the distributions should be isolated instead of all the functions. Because the logic for getting the data is different.
In that sense, we can write a different helper that generates data with its own kwargs as well.

notebooks/working/3.0-mab-independent-variable-simulation.md
bouzaghrane: done!

notebooks/working/3.0-mab-independent-variable-simulation.md
bouzaghrane: Done!

notebooks/working/3.0-mab-independent-variable-simulation.md
timothyb0912: Yeah, in general, `tour_id` can be the same for multiple trips on the same tour. However, if I remember correctly, this dataset has been filtered such that there is one tour for going to work/school and one tour returning from work/school. Some exploratory data analysis may answer the question here! Let me know if, when sampling to choice==1, you see individuals with multiple rows of the same `tour_id`. I think you won't due to the pre-processing that occurred.

timothyb0912:
@bouzaghrane Am I correct in thinking this branch can be closed?

bouzaghrane:
@timothyb0912 yes. Just closed it.

PR #101, opened 2020-05-03
Ho applying deconfounder on sim data

hassan-obeid:
Not very organized. We'll discuss over the call.


notebooks/working/ho_deconfounder_on_sim_data.py
```diff
@@ -0,0 +1,811 @@
...
+from functools import reduce
+
+
+
+import os
+os.listdir('.')
+# -
+
+PATH = '../../data/raw/'
+file_name = 'simulated_long_format_bike_data.csv'
```
timothyb0912, line 32: @bouzaghrane can you check in the code where you create and save this dataset? We really shouldn't have any data files created mysteriously.

notebooks/working/ho_deconfounder_on_sim_data.py
```diff
@@ -0,0 +1,811 @@
...
+
+data = pd.read_csv(PATH+file_name)
+data = data.drop('Unnamed: 0', axis = 1)
```
timothyb0912, line 35: If you edit the code where one saves `simulated_long_format_bike_data`, you can do  `simulated_long_format_bike_data.to_csv(PATH+file_name, index=False)` and then you won't have the index saved as 'Unnamed: 0'.

notebooks/working/ho_deconfounder_on_sim_data.py
```diff
@@ -0,0 +1,811 @@
...
+             'cross_bay', 'household_size', 'num_kids', 
+              'cars_per_licensed_drivers', 
+             'gender'
+             
+            ]
+
+y_column = data['mode_id']
+
+# +
+sprinkler = CausalGraphicalModel(
```
timothyb0912, line 52: Could we use a more relevant name than sprinkler? I know that's the name used in the example for `CausalGraphicalModels` but it won't make any sense to readers of the notebook. For example, something like `causal_graph`?

notebooks/working/ho_deconfounder_on_sim_data.py
```diff
@@ -0,0 +1,811 @@
...
+# # mnl_names["cross_bay"] = ["Cross-Bay Tour (Shared Ride 2 & 3+)",
+# #                           "Cross-Bay Tour (All Transit Modes)"]
+# mnl_specification["cross_bay"] = [[2, 3]]
+# mnl_names["cross_bay"] = ["Cross-Bay Tour (Shared Ride 2 & 3+)"]
+
+# mnl_specification["household_size"] = [[2, 3]]
+# mnl_names["household_size"] = ['Household Size (Shared Ride 2 & 3+)']
+
+# mnl_specification["num_kids"] = [[2, 3]]
+# mnl_names["num_kids"] = ["Number of Kids in Household (Shared Ride 2 & 3+)"]
```
timothyb0912, line 141: I assume there's no reason to keep this completely commented out cell around? If so, could we remove it so that the notebook is easier to read / only contains what's necessary to understand what is happening?

notebooks/working/ho_deconfounder_on_sim_data.py
```diff
@@ -0,0 +1,811 @@
...
+
+import matplotlib
+matplotlib.rcParams.update({'font.sans-serif' : 'Helvetica',
+                            'axes.labelsize': 10,
+                            'xtick.labelsize' : 6,
+                            'ytick.labelsize' : 6,
+                            'axes.titlesize' : 10})
+import matplotlib.pyplot as plt
+
+import seaborn as sns
```
timothyb0912, line 231: Can we consolidate all the imports at the top of the notebook? Also, I imagine there are many imports that are simply copied here but not actually used. For example, LogisticRegression, roc_auc_score, etc.

Can we remove all the unused imports? The idea is to not include extraneous code that may distract or confuse readers of the notebook.

notebooks/working/ho_deconfounder_on_sim_data.py
```diff
@@ -0,0 +1,811 @@
...
+             'cross_bay', 'household_size', 'num_kids', 
+              'cars_per_licensed_drivers', 
+             'gender'
+             
+            ]
+
+y_column = data['mode_id']
+# -
+
+spec_dic = specifications(mnl_specification=mnl_specification, num_modes=8)
```
timothyb0912, line 254: What does `specifications` actually do? A comment in the function definition says "# Iterate over all the items in dictionary and filter items which has even keys" but there is no actual check for if the key is even. I believe the comment is outdated.

In general I would suggest two things.
1. Add docstrings to one's functions so readers know what the function is supposed to do. Currently people have to read every detail of the function to know what it they are doing.
2. Define functions before you use them. This allows someone to execute the notebook without error. Currently, a user who tries to `Run All` on this notebook would immediately get an error about undefined functions. If one wants to make the notebook easy to read by putting function definitions out of site, then I'd suggest putting the function definitions in a .py file that is imported.

notebooks/working/ho_deconfounder_on_sim_data.py
```diff
@@ -0,0 +1,811 @@
...
+spec_dic
```
timothyb0912, line 255: This spec_dict is completely opposite of the specification dictionary used in pylogit where each key is the variable, and each value is either (i) an alternative with its own coefficient for that variable or (ii) a set of alternatives that share a common coefficient for that variable.

What's the purpose of your `spec_dict`? Is it merely to understand what variables are in the utility equation of each alternative?

timothyb0912: Beyond even clarifying what `spec_dict` is for via github discussion, you should include a comment saying for what this object is for so that the readers of the notebook can understand. At the moment it is strange to need to call a `specifications` function on an object that is already called `mnl_specification`.

notebooks/working/ho_deconfounder_on_sim_data.py
```diff
@@ -0,0 +1,811 @@
...
+holdout_dfs = []
+masks_df = []
+rows_df =[]
+latent_dim = 1
+for i in data['mode_id'].unique():
+
+    data_mode_i = data[data['mode_id']==i]
+    # standardize the data for PPCA
+    print("Analysis for mode: ", i)
+    print("-------------------------------------------------------------------------------------------")
```
timothyb0912, line 267: The following should accomplish the same thing without taking up multiple lines. Feel free to use whatever number you want, 40 just seemed like a big enough value.
```suggestion
    print("-" * 40)
```

notebooks/working/ho_deconfounder_on_sim_data.py
```diff
@@ -0,0 +1,811 @@
...
+    X_columns = spec_dic[i]
+    
+    if 'intercept' in X_columns:
+        X_columns.remove('intercept')
+        print(i, X_columns)
+    
+    X = np.array((data_mode_i[X_columns] - data_mode_i[X_columns].mean())/data_mode_i[X_columns].std())
+    
+    confounders, holdouts, holdoutmasks, holdoutrow= confounder(holdout_portion=0.2, X=X, latent_dim=latent_dim)
```
timothyb0912, line 276: I would put the `holdout_portion` in a variable of its own that is defined outside the for-loop, the same way and for the same reasons as defining `latent_dim` outside the for-loop.

timothyb0912: Also, can you please label the plot axes for the plots being produced on each iteration of the for-loop. I have no idea what the plot is showing and why its there.

notebooks/working/ho_deconfounder_on_sim_data.py
```diff
@@ -0,0 +1,811 @@
...
+
+data['confounder_all_2'] = confounders_using_all[2]
+
+# ## Estimate original MNL
+
+# data
+
+
+# +
+# Estimate the basic MNL model, using the hessian and newton-conjugate gradient
```
timothyb0912, line 391: Below, you actually use `method='BFGS"`
```suggestion
# Estimate the basic MNL model, using the hessian and BFGS
```

notebooks/working/ho_deconfounder_on_sim_data.py
```diff
@@ -0,0 +1,811 @@
...
+mnl_model = cm.create_choice_model(data=data,
+                                   alt_id_col="mode_id",
+                                   obs_id_col="observation_id",
+                                   choice_col="sim_choice",
+                                   specification=mnl_specification,
+                                   model_type="MNL",
+                                   names=mnl_names)
+
+num_vars = len(reduce(lambda x, y: x + y, mnl_names.values()))
+# Note newton-cg used to ensure convergence to a point where gradient
```
timothyb0912, line 401:
```suggestion
# Note BFGS used to ensure convergence to a point where gradient 
```

notebooks/working/ho_deconfounder_on_sim_data.py
```diff
@@ -0,0 +1,811 @@
...
+# +
+# Create my specification and variable names for the basic MNL model
+# NOTE: - Keys should be variables within the long format dataframe.
+#         The sole exception to this is the "intercept" key.
+#       - For the specification dictionary, the values should be lists
+#         or lists of lists. Within a list, or within the inner-most
+#         list should be the alternative ID's of the alternative whose
+#         utility specification the explanatory variable is entering.
+
+mnl_specification_noncausal = OrderedDict()
```
timothyb0912, line 421: The following might be much clearer for readers than recreating the specification dictionary from scratch.
```suggestion
mnl_specification_noncausal = deepcopy(mnl_specification)
del mnl_specification_noncausal['total_travel_distance']
```

Note you may need to do `from copy import deepcopy`.

notebooks/working/ho_deconfounder_on_sim_data.py
```diff
@@ -0,0 +1,811 @@
...
+mnl_names_noncausal = OrderedDict()
```
timothyb0912, line 422: The following might be much clearer for readers than recreating the name dictionary from scratch.
```suggestion
mnl_names_noncausal = deepcopy(mnl_names)
del mnl_names_noncausal['total_travel_distance']
```

Note you may need to do `from copy import deepcopy`.

notebooks/working/ho_deconfounder_on_sim_data.py
```diff
@@ -0,0 +1,811 @@
...
+#     name: python3
+# ---
+
+# +
+import pandas as pd
+import numpy as np
+import matplotlib.pyplot as plt
+from causalgraphicalmodels import CausalGraphicalModel, StructuralCausalModel
+from collections import OrderedDict
+import pylogit as cm
```
timothyb0912, line 22: Look at you. I think I've only ever done `import pylogit as pl`! 😆

notebooks/working/ho_deconfounder_on_sim_data.py
```diff
@@ -0,0 +1,811 @@
...
+mnl_names_noncausal["cross_bay"] = ["Cross-Bay Tour (Shared Ride 2 & 3+)"]
+
+mnl_specification_noncausal["household_size"] = [[2, 3]]
+mnl_names_noncausal["household_size"] = ['Household Size (Shared Ride 2 & 3+)']
+
+mnl_specification_noncausal["num_kids"] = [[2, 3]]
+mnl_names_noncausal["num_kids"] = ["Number of Kids in Household (Shared Ride 2 & 3+)"]
+
+# +
+# Estimate the basic MNL model, using the hessian and newton-conjugate gradient
```
timothyb0912, line 473:
```suggestion
# Estimate the basic MNL model, using the hessian and BFGS
```

notebooks/working/ho_deconfounder_on_sim_data.py
```diff
@@ -0,0 +1,811 @@
...
+mnl_model_noncausal = cm.create_choice_model(data=data,
+                                   alt_id_col="mode_id",
+                                   obs_id_col="observation_id",
+                                   choice_col="sim_choice",
+                                   specification=mnl_specification_noncausal,
+                                   model_type="MNL",
+                                   names=mnl_names_noncausal)
+
+num_vars_noncausal = len(reduce(lambda x, y: x + y, mnl_names_noncausal.values()))
+# Note newton-cg used to ensure convergence to a point where gradient
```
timothyb0912, line 483:
```suggestion
# Note BFGS used to ensure convergence to a point where gradient 
```

notebooks/working/ho_deconfounder_on_sim_data.py
```diff
@@ -0,0 +1,811 @@
...
+# Note newton-cg used to ensure convergence to a point where gradient 
+# is essentially zero for all dimensions. 
+mnl_model_causal.fit_mle(np.zeros(num_vars),
+                  method="BFGS")
+
+# Look at the estimation results
+mnl_model_causal.get_statsmodels_summary()
+# -
+
+# ## Estimate Causal MNL - method 2
```
timothyb0912, line 523: Can you add a description in text to explain the differences between your various methods of estimating a causal MNL?

notebooks/working/ho_deconfounder_on_sim_data.py
```diff
@@ -0,0 +1,811 @@
...
+    
+#     print(len(data.loc[data['mode_id']==i, col_name]), len(confounder_vectors[int(i-1)][2]) )
+    
+    col_name = 'confounder_for_mode_' + str(int(i))
+    data.loc[data['mode_id']==i, col_name] = confounder_vectors[int(i-1)][2]
+    data[col_name] = data[col_name].fillna(0)
+    
+data['confounder_all'] = data[['confounder_for_mode_1','confounder_for_mode_2','confounder_for_mode_3',
+                              'confounder_for_mode_4', 'confounder_for_mode_5', 'confounder_for_mode_6',
+                              'confounder_for_mode_7', 'confounder_for_mode_8']].sum(axis=1)
```
timothyb0912, line 380: I don't at all understand why you are summing the inferred confounders for each mode.

Shouldn't use simply use each modes inferred confounder, separate from the inferred confounder of other modes?

To be clearer, the confounding variable is travel distance, which can differ across modes. For instance, the travel distance by bike and the travel distance by car are different because an individual may drive on highways but they can only bike on local roads. Summing the confounders is conceptually equivalent to summing the travel distances, and I don't think one should do that.

notebooks/working/ho_deconfounder_on_sim_data.py
```diff
@@ -0,0 +1,811 @@
...
+latent_dim = 1
+for i in data['mode_id'].unique():
+
+    data_mode_i = data[data['mode_id']==i]
+    # standardize the data for PPCA
+    print("Analysis for mode: ", i)
+    print("-------------------------------------------------------------------------------------------")
+    X_columns = spec_dic[i]
+    
+    if 'intercept' in X_columns:
```
timothyb0912, line 270: I believe this cell has a major problem. It only removes the intercept from the values used to estimate the confounder.

Problematically, total travel distance (i.e. the confounding variable) is still left in the set of columns used to infer total travel distance.

notebooks/working/ho_deconfounder_on_sim_data.py
```diff
@@ -0,0 +1,811 @@
...
+data.loc[data['mode_id']==1, 'confounder_da'] = confounders_DA[2]
+data['confounder_da'] = data['confounder_da'].fillna(0)
+
+data.loc[data['mode_id']==2, 'confounder_sr2'] = confounders_SR2[2]
+data['confounder_sr2'] = data['confounder_sr2'].fillna(0)
+
+data.loc[data['mode_id']==3, 'confounder_sr3'] = confounders_SR3[2]
+data['confounder_sr3'] = data['confounder_sr3'].fillna(0)
+
+data['confounder_all_2'] = data[['confounder_da', 'confounder_sr2', 'confounder_sr3']].sum(axis=1)
```
timothyb0912, line 556: Again, why are you summing the confounders across travel modes?

notebooks/working/ho_deconfounder_on_sim_data.py
```diff
@@ -0,0 +1,811 @@
...
+
+# +
+data_mode_specific = data[data['mode_id']==1]
+
+data_mode_specific.plot.scatter('total_travel_distance', 'confounder_all')
+
+
+# -
+
+# ## Putting everything in a function
```
timothyb0912, line 607: Before you conclude the notebook, I think you should have a final cell where you answer "So what?"

In other words, what are the high-level takeaways from using the deconfounder on the simulated data? Did the estimated coefficients have lower bias when using the inferred confounders, relative to simply omitting travel distance? I would include a dataframe that shows:
- the true coefficients used to simulate the data.
- estimated coefficients with all variables (including total travel distance)
- estimated coefficients without total travel distance
- estimated coefficients with inferred confounders using your 1st method
- estimated coefficients with inferred confounders using your 2nd method

Additionally, just for clarity I would make and display a second dataframe that shows the absolute bias in the estimated coefficients, as in take the absolute value of [the first dataframe - the column for the true coefficients used to simulate the data].

timothyb0912 (changes requested):
Okay, overall, I would focus on the following things:
1. Clearly explaining why one is doing everything that is being done. This would come in the form of comments explaining what one is doing on most lines, docstrings explaining what is being done in a function + its inputs and outputs, and text explaining what one is doing in each section of the notebook.
2. Making sure the notebook is completely runnable. One should be able to hit `run all` and have the notebook work.
3. Making sure that one's inference for the confounder makes sense. In particular either removing or justifying the summing of the inferred confounders across travel modes.

notebooks/working/ho_deconfounder_on_sim_data.py
bouzaghrane: Will do!

notebooks/working/ho_deconfounder_on_sim_data.py
bouzaghrane: or name_of_alt_causal_graph

notebooks/working/investigating_deconfounder.py
```diff
@@ -79,6 +78,7 @@
 # -
 
 # ## Specify the assumed causal graph of the data generation process
+# The causal graph below represents the true data generating process that we'll be using for simulation. Notice how the confounder affects variables a, b, d, e, and f, as well as the outcome y. 
 
 # +
 data_generation_graph = CausalGraphicalModel(
```
bouzaghrane, line 84: Might not be necessary, but I would just move the nodes and edges lists to be separate parameters at the top of the notebook.

notebooks/working/investigating_deconfounder.py
```diff
@@ -114,35 +114,54 @@
...
+f = -.3*confounder + np.random.normal(loc=-12,scale=3,size=sample_size)
 
+# +
+## Generating coefficients
+beta_a = np.random.normal(loc=2, scale=1, size=sample_size)
+beta_b = np.random.normal(loc=2, scale=1, size=sample_size)
+beta_c = np.random.normal(loc=-5, scale=1, size=sample_size)
+beta_d = np.random.normal(loc=3, scale=1, size=sample_size)
+beta_e = np.random.normal(loc=2, scale=1, size=sample_size)
+beta_f = np.random.normal(loc=-5, scale=1, size=sample_size)
```
bouzaghrane, line 145: Maybe not necessary, but you can have one dictionary storing the mean and std of each of these parameters.

notebooks/working/investigating_deconfounder.py
```diff
@@ -114,35 +114,54 @@
...
 
+beta_confounder = np.random.normal(loc=7, scale=1, size=sample_size)
+
+# +
 df = pd.DataFrame(np.array([a,b,c,d,e,f, confounder]).T, columns = ['a', 'b', 'c',
```
bouzaghrane, line 150: small pep8 work. all will be fixed when linting.

notebooks/working/investigating_deconfounder.py
```diff
@@ -151,29 +170,39 @@
...
+#
+# How do we go about checking for whether a confounder exists (in this example, we know that a confounder exist since we defined the data generating process, but in the real world, the data generating process is more often than now unknown to the analyst). 
+#
+# There are a multitude of tests and ways to go about answering this question. A simple, off-the-shelf check is the bartlett sphericity test, which essentially just compares the correlation matrix of our covariates to the identity matrix. This test is used to check whether our covariate space can be reduced to a lower-dimensional space of factors that explain the variation in our data. 
+#
+# ### Tim, this would be a good place to add your other tests. I'm thinking of the bullet points below:
+# - We can hypothesize a causal graph, and run conditional independence tests. 
+# - Prior predictive checks? 
 
 from factor_analyzer.factor_analyzer import calculate_bartlett_sphericity,calculate_kmo
```
bouzaghrane, line 207: All imports at the top of the notebook.

notebooks/working/investigating_deconfounder.py
```diff
@@ -187,74 +216,30 @@
...
 
 # +
 latent_dim = 1
 
-confounders, holdouts, holdoutmasks, holdoutrow= confounder_ppca(holdout_portion=0.2, X=X, latent_dim=latent_dim)
+confounders, holdouts, holdoutmasks, holdoutrow= confounder_ppca(holdout_portion=0.2, X=X, 
+                                                                 latent_dim=latent_dim, linear = False)
 
 holdouts_req = [holdouts]
 display.display(len(holdouts_req))
```
bouzaghrane, line 245: I would just describe using in-line comments what each of the loops below does.

notebooks/working/investigating_deconfounder.py
```diff
@@ -532,16 +568,25 @@ def confounder_ppca(X, latent_dim, holdout_portion):
...
                     scale=stddv_datapoints * tf.ones([num_datapoints, data_dim]),
                     name="x")  # (modeled) data
+        else: 
+            x = ed.Normal(loc=tf.multiply(tf.matmul(z, w), 1-holdout_mask) 
+                          + tf.multiply(tf.matmul(z**2, w**2), 1-holdout_mask),
+                    scale=stddv_datapoints * tf.ones([num_datapoints, data_dim]),
+                    name="x")
+            
+            
         return x, (w, z)
```
bouzaghrane, line 590: Can this go outside of the `confounder_ppca` function? And then just call it directly inside the other function.

bouzaghrane, line 590: Can this go outside of the confounder_ppca function? And then just call it directly inside the other function.

notebooks/working/investigating_deconfounder.py
```diff
@@ -638,39 +684,28 @@ def interceptor(rv_constructor, *rv_args, **rv_kwargs):
...
+                      + tf.multiply(tf.matmul(z**2, w**2), 1-holdout_mask),
+                scale=stddv_datapoints * tf.ones([num_datapoints, data_dim]),
+                name="x")
 
-def variational_model(qw_mean, qw_stddv, qz_mean, qz_stddv):
-    qw = ed.Normal(loc=qw_mean, scale=qw_stddv, name="qw")
-    qz = ed.Normal(loc=qz_mean, scale=qz_stddv, name="qz")
-    return qw, qz
 
+    return x, (w, z)
```
bouzaghrane, line 705: Ah just saw it here. Then feel free to just remove its definition from inside `confounder_ppca`

bouzaghrane (changes requested):
Hassan, In addition to the comments Tim had in the code, I think this would just need getting the following:
 - Avoid nesting definition of functions.
 - Run flake 8 to just have pep compliant code.
 - Add in-line comments as necessary to make anything abstract clearer
 - Add documentation to your functions

Let me know if I can help with this, brother.

PR #109, opened 2020-05-14
Example of potential usefulness of causal graphs in context of recent paper.

timothyb0912:
@bouzaghrane and @hassanobeid1994 please take a look at the notebook added in this pull request.

I think this represents a good example of where causal diagrams and the general causal inference workflow would have been useful in a recent paper out of Ben-Akiva's research group.

Let me know what you all think!


notebooks/working/_04-tb-parking-choice-example.py
```diff
@@ -0,0 +1,101 @@
...
+import sys
+
+sys.path.insert(0, '../../src/')
+from graphs import parking
+# -
+
+causal_graph = parking.PARKING_CAUSAL_MODEL.draw()
+causal_graph.graph_attr.update(size="10,6")
+causal_graph
+
```
bouzaghrane, line 38: Should we add a description of all the variables in the graph?

bouzaghrane (commented):
This looks great to me Tim!

notebooks/working/_04-tb-parking-choice-example.py
timothyb0912: It feels like a lot of work (or maybe I'm just lazy). At the least, I should include a line in the module docstring that tells readers to check the paper for definitions of all variables.

PR #126, opened 2020-06-16
PC Algorithm investigation

timothyb0912:
Testing out the PC algorithm on the variables from our drive alone utility's causal graph.


timothyb0912:
Please take a look @bouzaghrane and @hassanobeid1994 .

I included the written description of the PC Algorithm from its original authors, followed the steps and showed the causal graph at each iteration, and ended with some conclusions and comparison to the causal graph that we started with.

Overall, I don't think it was too tedious. For a PhD student or full-time researcher, this effort could be repeated for all utility functions and still be completed within a week.

PR #154, opened 2020-11-26
Add discussion sec

timothyb0912:
# What
This PR adds a discussion section to the end of the paper, before the conclusion section.

Please leave comments on `article/sections/_8_discussion.tex`. 🙏 

Other changes include updating the bibliography, updating the `main.tex` file so the discussion shows up on the final pdf, and renaming the conclusion section.


article/sections/_8_discussion.tex
```diff
@@ -0,0 +1,86 @@
...
+Our latent confounding example showed that statistical inference of our models' parameters may still be a challenge, even with a correct causal graph.
+Such challenges lead us to the following set of post-graph-construction topics:
+\begin{itemize}
+   \item model estimation
+   \item model checking
+   \item experimental design
+   \item experiment analysis
+   \item decision analysis
+\end{itemize}
+To us, each item above is important for getting credible results from our analysis and for maximizing our interventions' benefits.
```
bouzaghrane, line 14: I'm a bit confused by how we will maximize our interventions' benefits just from our statistical inference? I think it might be best to say something like...

article/sections/_8_discussion.tex
```diff
@@ -0,0 +1,86 @@
...
+Accordingly, even though we will neglect details due to space and time constraints, we briefly discuss these topics below.
+
+Perhaps most obviously, after creating and testing a causal graph, we will use it to estimate the effects of our interventions.
+To compute our effect estimates, we will need to evaluate statements such as the probability of a particular node's variable taking a given value, conditional on the values of that node's parents.
```
bouzaghrane:
```suggestion
To compute our effect estimates, we will need to evaluate statements such as the probability of a particular node taking a given value, conditional on the values of that node's parents.
```

article/sections/_8_discussion.tex
```diff
@@ -0,0 +1,86 @@
...
+These probabilities will come from our estimated models, so model estimation is critically important to our effect estimation.
+Thankfully, this is the part of causal inference that choice \textit{modellers} are most familiar with.
+For instance, \citet{kostic_2020_uncovering} first use the PC and GES causal discovery algorithms to generate a causal graph, then they test the graph qualitatively, and finally they estimate models corresponding to this causal graph.
```
bouzaghrane, line 21:
```suggestion
For instance, \citet{kostic_2020_uncovering} first use the PC and GES causal discovery algorithms to generate a causal graph, then test the graph qualitatively, and finally estimate models corresponding to this causal graph.
```

article/sections/_8_discussion.tex
```diff
@@ -0,0 +1,86 @@
...
+As another example, \citet{garrido_2020_estimating} start from where we end: at a known (or selected) causal graph.
+They then use neural network density estimators to model the necessary probabilities for estimating one's causal effects.
```
bouzaghrane:
```suggestion
They then use neural network density estimators to model the necessary probabilities for estimating one's causal effects of interest.
```

article/sections/_8_discussion.tex
```diff
@@ -0,0 +1,86 @@
...
+
+Next, after estimating the models for our causal graph but before interpreting or using our results, we should check our entire system of models.
+Here, there are multiple, non-competing ways of performing these diagnostics.
+We can check our models separately, jointly, or in subsets.
+And if we're being thorough, we can perform all these checks instead of one kind.
```
bouzaghrane:
```suggestion
For thoroughness' sake, we can perform all these checks instead of one kind.
```

article/sections/_8_discussion.tex
```diff
@@ -0,0 +1,86 @@
...
+
+From a disaggregate perspective, we can consider a sequential application of model checking exercises, one per estimated model.
+Hopefully, each model checking process will include the use of visual diagnostics, as (for example) described in \citet{brathwaite_2018_check}.
```
bouzaghrane:
```suggestion
Ideally, each model checking process will include the use of visual diagnostics, as (for example) described in \citet{brathwaite_2018_check}.
```

article/sections/_8_discussion.tex
```diff
@@ -0,0 +1,86 @@
...
+Alternatively, we can follow a subset approach used by \citet{tran_2016_model}.
+There, we jointly check the models for all variables that our intervention will set (i.e. the treatment assignment variables), and then we separately check the model(s) for our outcome variable(s).
+Generally, we can check subsets of models together instead of checking one model at a time.
```
bouzaghrane: This is a duplicate in my opinion. I suggest deleting it or merging it with line 32 as follows:

Alternatively, we can check subsets of models together instead of checking one model at a time, as shown by \citet{tran_2016_model}.

article/sections/_8_discussion.tex
```diff
@@ -0,0 +1,86 @@
...
+Finally, as in \citet{williamson_2013_model}, we may define global diagnostic measures over all nodes in our causal graph.
+This corresponds to checking all our models jointly, and it represents yet another diagnostic approach.
```
bouzaghrane:
```suggestion
Finally,  we can check all our models jointly by defining global diagnostic measures over all nodes in our causal graph, as demonstrated by \citet{williamson_2013_model}.
```

article/sections/_8_discussion.tex
```diff
@@ -0,0 +1,86 @@
...
+
+Following model diagnostics, we are ready to use our models and causal graph to inform real interventions.
+These interventions can come in two kind: an experiment or a ``full-scale'' implementation of one's policy.
+If we are setting up an experiment, then we are typically interested in one of two aims.
```
bouzaghrane:
```suggestion
If our intervention is experimental, then we are likely interested in one of two aims.
```

article/sections/_8_discussion.tex
```diff
@@ -0,0 +1,86 @@
...
+We either want to decide between one or more treatment options, or we want to learn about our system, though not necessarily to make a decision.
+In both cases, however, we pay great attention to the design of our experiments.
+
+When experimenting to make decisions, such as whether to launch a given treatment or not, we pay extra attention to the size of our experiment.
+Specifically, we want our experiment's sample size to be large enough such that after we update our beliefs using the experimental data, that we have at least our minimum desired probability of making the correct decision.
```
bouzaghrane, line 44: Can we just say power here instead of minimum desired probability of making the correct decision?

article/sections/_8_discussion.tex
```diff
@@ -0,0 +1,86 @@
...
+The decision can be to declare the effect of a treatment statistically different from zero, but more frequently, the decision will be more fundamental such as "implement treatment A."
+For thorough explanations of how to conceptualize and design experiments in a bayesian, model-based setting, see \citet{chaloner_1995_bayesian} and \citet{wang_2002_simulation}.
+For examples and guidance on how to use one's causal graph structure to guide the general design of one's experiment, beyond sample size, see \citet{madrigal_2007_cluster}.
+There, the structure of one's causal graph is used to inform general design decisions such as the clustered allocation of individuals to treatment, and the experimental design is itself analyzed graphically.
```
bouzaghrane, line 48: Can you explain the difference between the first and second portions of this sentence? Seems like the first portion is just a portion of experimental design.

article/sections/_8_discussion.tex
```diff
@@ -0,0 +1,86 @@
...
+Additionally, note the relations to reinforcement learning where an agent has to perform experiments in order to discover the action that will maximize her reward.
+There, \citet{lee_2018_structural} have shown that designing our experiments without guidance from one's causal model is generally suboptimal, and that we can achieve optimality by leveraging our causal graph to design our experimentation plan.
```
bouzaghrane: I don't understand how these two sentences relate to each other. Reinforcement learning here sounds like "try and try until you get the reward you want". The second sentence sounds like it's more informed. I understand how the causal model can be used to model an output reward for an agent but I am not fully sure how that relates to RL.

Maybe I'm just missing something here.

article/sections/_8_discussion.tex
```diff
@@ -0,0 +1,86 @@
...
+
+Now, let's transition from experimentation for decision making to consider experimentation for learning.
+Imagine that we are at a transportation network company and that we are running a pricing experiment to learn about price elasticities of our customers.
+Here, there is no immediate decision being made, but we learn about an edge in our causal graph: the edge from price of a trip (treatment) to purchase of the ride (outcome).
+In other cases, we may experiment to learn not just about the strength of an edge, but about the presence of edges and the structure of the graph more generally.
+For example, we may wish to remove residual ambiguity from a causal discovery process that output a Markov Equivalence Class of graphs instead of a single causal DAG.
```
bouzaghrane: Also, just want to make sure we're going to be using DAGs, causal graphs, causal DAGs, etc.. interchangeably. Is that the case? Do you also think it's necessary to mention that we will be doing just that in the into section?

```suggestion
For example, we may wish to remove residual ambiguity from a causal discovery process that outputs a Markov Equivalence Class of graphs instead of a single causal DAG.
```

article/sections/_8_discussion.tex
```diff
@@ -0,0 +1,86 @@
...
+In these situations, we're interested in optimally designing an experiment (or series of experiments) to learn a causal graph (or its properties), and we're interested in how we can leverage potentially mutliple experimental datasets to improve our causal graphs.
```
bouzaghrane:
```suggestion
In these situations, we are interested in optimally designing an experiment (or series of experiments) to learn a causal graph (or its properties). We are further interested in how we can leverage potentially multiple experimental datasets to improve our causal graphs.
```

article/sections/_8_discussion.tex
```diff
@@ -0,0 +1,86 @@
...
+For a review of the literature on experimentation for learning and construction/refinement of a preliminary causal graph, see \citet[Sec. 3.1.2]{kalisch_2014_causal}.
+For more recent approaches in this vein, see works such as \citet{triantafillou_2015_constraint, kocaoglu_2017_experimental, brouillard_2020_differentiable, rantanen_2020_learning}.
+
+Finally, after we have run any experiments that we are interested in, we still must decide what to do in our population.
```
bouzaghrane:
```suggestion
Finally, after we have run any experiments that we are interested in, we still must decide what to do in our population with respect to our intervention.
```

article/sections/_8_discussion.tex
```diff
@@ -0,0 +1,86 @@
...
+Three major types of questions come to mind immediately:
+\begin{enumerate}
+   \item Should we launch the treatment(s) at all?
+   \item Should we launch the treatments to everyone?
+   \item How should the treatments be dispersed/implemented?
+\end{enumerate}
+
+After updating our posterior beliefs with the experimental data, we'll want to analyze and come to a conclusion about launching our treatment(s).
+In particular, we will wish to determine the expected distribution of impacts under each scenario (launch treatment(s), business as usual, etc.).
```
bouzaghrane:
```suggestion
In particular, we will wish to determine the expected distribution of impacts under each alternative decision about the treatment.
```

article/sections/_8_discussion.tex
```diff
@@ -0,0 +1,86 @@
...
+Here, \citet{manski_2019_treatment} should provide the basic idioms of thought and pointers to the larger literature on treatment choice from a decision theoretic perspective.
+
+Next, if we've decided to launch the treatments, we come to the question of who should receive the treatment(s)?
+Everyone?
+A select few?
+Are there certain subgroups that should receive the treatment but not others?
```
bouzaghrane, line 77: Are you going to itemize these like you did above?

bouzaghrane: I also feel like there are more questions to consider than just these and the ones above as well. These questions will of course be based on the context at hand, so it might make sense to add a short sentence saying that additional considerations could be needed depending on the context of interest. thoughts?

bouzaghrane (changes requested):
@timothyb0912 here are my comments for the discussion section.

article/sections/_8_discussion.tex
timothyb0912: I don't know, it seems grammatically incorrect to not have exactly the same sentence structure in each part of the parallel syntax: subject-action, subject-action, subject-action. Your suggestion removes the subject from elements 2 and 3 in the list but not for the first. See here for what I'm thinking: https://owl.purdue.edu/owl/general_writing/mechanics/parallel_structure.html

article/sections/_8_discussion.tex
timothyb0912: I don't think so. I don't think these two concepts are the same.

Power is the probability of correctly deciding that an alternative hypothesis is true, when the alternative hypothesis is true, in a hypothesis testing context.

I'm actually not operating in an hypothesis testing context at all. Imagine you're a ridesharing company doing an A/B test to select amongst one of two candidate models of whether a person buys a ride given a  particular fare quote. I am not deciding if one model is true or the other is false. I know both are likely false. I'm wondering which model's usage leads to greater revenue/profit. I'm worried about choosing incorrectly and suffering opportunity costs. If model B will perform better than model A (even if both are false) then I don't want to select model A, and I want to control the probability that sampling variability could lead me to mistakenly select model A.

timothyb0912: In general, power and type I error and those sorts of terms are about null hypothesis significance testing (NHST).

I'm talking about decision theory, which is much broader than NHST.

article/sections/_8_discussion.tex
timothyb0912: I wasn't planning on itemizing them. Do you think the bullets would be useful?

And do you have some examples of other questions you think would come up for consideration? Maybe an example project / context in which the questions would arise?

article/sections/_8_discussion.tex
timothyb0912: Hmm, there's a lot in your comment. First, note that "try and try" = "experiment and experiment." An agent taking an action IS intervening on their system (i.e., applying the do-operator). This addresses your second and third sentences that seems to draw a sharp distinction between RL and causal inference: "RL here sounds like [...]. The second sentence sounds like it's more informed."

Second the relation is that the cited paper is 100% focused on about multi-armed bandit problems, and that is a classic / one of the oldest reinforcement learning problems: https://en.wikipedia.org/wiki/Multi-armed_bandit. This addresses your second to last sentence "I am not fully sure how that relates to RL."

Lastly, the relation to RL is that initially the agent starts out unaware of the causal graph of the system they operate in, and the agent has to either observe others or actively experiment/intervent/act on the system to learn about the graph. This is the same as my describing to a human agent that they have to experiment on their system to learn about their causal graph.

I made the following changes. Does that make it clearer?
```suggestion
Additionally, note the relations to reinforcement learning where an agent has to perform experiments in order to discover the action/intervention that will maximize her expected, counterfactual reward.
In this context, \citet{lee_2018_structural} have shown that designing our experiments without guidance from one's causal model is generally suboptimal, and that we can achieve optimality by leveraging our causal graph to design our experimentation plan.
```

article/sections/_8_discussion.tex
timothyb0912: Say I'm at a ridesharing company. We run an experiment to test a new model for selecting the pickup spot for passengers. As part of this experiment, we collect data on metrics: revenue, pick up times, cancellation rate, driver-passenger contact rates, etc. The effects on each of these metrics that we track are the treatment effects. The decision is whether or not to implement the new model though. In the best case scenario, we'd have a utility function that allows us to combine how much we care about of the various attributes (aka metrics), and we'd solve the choice modelling problem of utility maximizing.

Do you see the difference now? Making inference on attributes (i.e. metrics, i.e. effects on specific outcome variables) vs making inference on the full decision, i.e. the weighting across these metrics / effects.

article/sections/_8_discussion.tex
timothyb0912: I'm curious about the rest of your sentence!

In general though, I'm with you! Please note that the word "just" is not at all in my sentence. 😝

PR #198, opened 2020-12-21
Pdf for review

timothyb0912:
# What
This is the semi-final pdf for review. The following items are still to be added:
- Author listing pending discussion (see #199).
- Brief section 2 edit to address #195.

cc @hassan-obeid and @bouzaghrane for visibility.

PR #200, opened 2020-12-21
Final Edits for Section2

### LINKED ISSUE #195: Strengthen Section 2 results declaration
# What
The current simulation [results](https://github.com/hassan-obeid/tr_b_causal_2020/blob/master/notebooks/final/7.0-mab-selection-on-observables-final.ipynb) indicate a small (but non-trivial) probability of a positive treatment effect (i.e. more drivers) due to moving people closer to work.
This would mean the time and cost reductions for driving out-weighted the closer distance.

We should change the statement of our results in the article to highlight the fact that
our simulations show an example where using an incorrect causal graph can
cause one to not just have the wrong magnitude of interred treatment effect but also the wrong sign.

cc @hassan-obeid and @bouzaghrane for visibility.

bouzaghrane:
This PR follows up on #198 with final edits to section 2 about the conclusions of the selection on observables simulation.

### DIFF
--- a/article/sections/_2_graph_importance.tex
+++ b/article/sections/_2_graph_importance.tex
@@ -207,6 +207,8 @@ aimed at reducing travel distance.
 These differences are plotted under the following assumptions: that the outcome model is the same in both scenarios and that the only difference between the two scenarios is the set of variables affected by the proposed intervention.
 This difference in the set of variables affected by the proposed intervention is a result of the two different causal graphs representing each scenario (Figures \ref{fig:IND_GRAPH} and \ref{fig:DA_causal_2}).
 Figure \ref{fig:histogram_probability} highlights the resulting bias between the estimated probability of an average individual choosing a car centric mode.
+The histograms show that the distribution of inferred treatment effects based on the ``true'' causal graph includes both positive and negative treatment effects.
+Using the wrong causal graph in one's analyses could result in not only wrong magnitudes of the treatment effect of interest but also the wrong sign.
 This plotted discrepancy shows the importance of using causal graphs in the estimation of treatment effects.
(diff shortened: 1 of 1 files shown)


article/sections/_2_graph_importance.tex
```diff
@@ -207,6 +207,8 @@ \section{Perils of disregard}
 These differences are plotted under the following assumptions: that the outcome model is the same in both scenarios and that the only difference between the two scenarios is the set of variables affected by the proposed intervention.
 This difference in the set of variables affected by the proposed intervention is a result of the two different causal graphs representing each scenario (Figures \ref{fig:IND_GRAPH} and \ref{fig:DA_causal_2}).
 Figure \ref{fig:histogram_probability} highlights the resulting bias between the estimated probability of an average individual choosing a car centric mode.
+The histograms show that the distribution of inferred treatment effects based on the ``true'' causal graph includes both positive and negative treatment effects.
+Using the wrong causal graph in one's analyses could result in not only wrong magnitudes of the treatment effect of interest but also the wrong sign.
```
timothyb0912: Sweet, this all looks good to me. Here's my one suggestion. I imagine someone pointing out that the probability of positive treatment effects is small, so I think it's better to pre-empt such discussion.
```suggestion
With low but non-trivial probability, using the wrong causal graph in one's analyses could result in not only wrong magnitudes of the treatment effect of interest but also the wrong sign.
```

timothyb0912 (commented):
👍🏽 to the changes. I left one small nitpick, but feel free to merge after addressing the main point.

bouzaghrane:
@timothyb0912 merged edit for section 2 #198

bouzaghrane:
Thanks for merging this @timothyb0912 !

PR #211, opened 2020-12-29
Latest edits of section 2

bouzaghrane:
# What

This PR covers the latest edits to section 2 based on the discussion from the meeting on Dec 28th, 2020. The following changes were made:
- Remove references to all non car-centric modes
- Reduce the description of the simulation procedures
- Remove non car-centric DAGs
- Update car-centric DAGs to show which variables are modified based on the data generating process assumed in each DAG

@hassan-obeid please take a look at the changes then pass it along to @timothyb0912


timothyb0912 (changes requested):
@bouzaghrane thanks for getting to this so quickly!

Right now, the new utility graphs that you added are pngs with no information about where they came from.

Can you make these images in graphviz?
In particular the process might look as follows.
For Figure "Independent_graph,"
1. See [here](https://graphviz.readthedocs.io/en/stable/examples.html?highlight=color#er-py) for an example of setting a node's or edge's color attribute.
2. Create a node for "proposed intervention," and color its outline white.
3. Draw an edge from the proposed intervention node to travel distance, and color it red.
4. Color the edge from travel distance to utility as red.
5. Color the node outline for utility as red.

Then, can you save the image as a pdf (better ability to resize as compared to png), and put it in the `article/images` directory by editing the script [here](https://github.com/hassan-obeid/tr_b_causal_2020/blob/master/workflow/store_sob_graphs.py)? You'd do a similar process for the DA_interacting_graph.

Let me know if any of that is unclear!

bouzaghrane:
@timothyb0912 please check the latest commit to make sure all is good. I added graphs as .py file in the ```graphs``` folder. The graphs now also show that all the affected nodes with a red outline.

workflow/store_sob_graphs.py
```diff
@@ -2,17 +2,18 @@
...
 """
-import causal2020.observables.graphs.sob as graphs
 import click
+from causal2020.graphs.sob import DA_IND_GRAPH, DA_INTERACTING_GRAPH,
+                                  SHARED_2_INTERACTING_GRAPH, SHARED_3P_INTERACTING_GRAPH
 from causal2020 import utils 
 
 
-def main_func(supplier=graphs, writer=utils) -> bool:
+def main_func(writer=utils) -> bool:
```
timothyb0912: Hey @bouzaghrane what's the purpose for this change?

I think it's bad to depend on global variables. It makes testing a real pain.
By removing the explicit injection via function argument, you make the function directly depend on the global variable.
When testing this means we'd now have to mock the variable and increase the complexity of the tests.

If there is no reason for the change, can you reset the HEAD of this branch it to the way it was before this commit?

timothyb0912 (changes requested):

workflow/store_sob_graphs.py
bouzaghrane: I reversed back in the latest commit.

bouzaghrane:
@timothyb0912 finally the checks have all passed. Let me know if you have any comments on how the code looks now.

article/sections/_2_graph_importance.tex
```diff
@@ -133,79 +121,21 @@ \section{Perils of disregard}
...
 Readers interested in exploring the details of our simulation exercise can refer to our GitHub repository \citep{brathwaite_etal_2020}.
 
 
-
 We then plot histograms of the computed differences between the average probability of an individual
 in our sample choosing a car centric mode before and after implementing a policy or intervention
 aimed at reducing travel distance.
 These differences are plotted under the following assumptions: that the outcome model is the same in both scenarios and that the only difference between the two scenarios is the set of variables affected by the proposed intervention.
-This difference in the set of variables affected by the proposed intervention is a result of the two different causal graphs representing each scenario (Figures \ref{fig:IND_GRAPH} and \ref{fig:DA_causal_2}).
+This difference in the set of variables affected by the proposed intervention is a result of the two different causal graphs representing each scenario (Figures \ref{fig:IND_GRAPH}, \ref{fig:DA_causal_2}, \ref{fig:SR2_causal_2}, and \ref{fig:SR3_causal_2}).
```
timothyb0912, line 138: Assuming there's no reason for using SR3_causal_2, I think you meant the following:
```suggestion
This difference in the set of variables affected by the proposed intervention is a result of the two different causal graphs representing each scenario (Figures \ref{fig:IND_GRAPH}, \ref{fig:DA_causal_2}, \ref{fig:SR2_causal_2}, and \ref{fig:DA_causal_2}).
```

timothyb0912 (approved):
I didn't take a look at the text, but the code seems fine to me!
Please re-request my review after @hassan-obeid is done looking at it.
Thanks again y'all.

Oh, @bouzaghrane could you add a commit with the updated pdf from running the entire project on your local machine?
You should just need to do `make article`, and then upload the pdf when it's done.

article/sections/_2_graph_importance.tex
```diff
@@ -44,23 +44,19 @@ \section{Perils of disregard}
 We model this travel mode choice problem based on a dataset from \citet{brathwaite_asymmetric}.
 This dataset is based on the 2012 California Household Travel Survey, and it
 contains approximately 4000 home-based school or work tours made by approximately 3850 individuals in the California Bay area.
-The dataset includes the following eight travel modes:
+The dataset includes eight travel modes. 
+For our illustrative purposes, we focus on the following car-centric modes and will not further present details about other modes:
```
hassan-obeid:
```suggestion
For our illustrative purposes, we focus on the following car-centric modes:
```

article/sections/_2_graph_importance.tex
```diff
@@ -44,23 +44,19 @@ \section{Perils of disregard}
...
    \item Shared Ride 3+: The individual shares an automobile ride with two or more individuals
-   \item Walk-Transit-Walk: The individual uses transit, walking to their transit origin and from their transit destination to the final destination
-   \item Drive-Transit-Walk: The individual uses transit, driving to access transit and walking from a transit stop to their final destination
-   \item Walk-Transit-Drive: The individual uses transit, walking to access transit and driving from a transit stop to their final destination
-   \item Bike: The individual uses a bike to make the trip
-   \item Walk: The individual walks to make the trip
 \end{itemize}
 
 Readers interested in a more detailed description of the dataset can refer to \citet{brathwaite_asymmetric}.
 For purposes of this exercise, we consider the Multinomial Logit model defined in \citet{brathwaite_asymmetric} as the true outcome generating model.
```
hassan-obeid:
```suggestion
For the purposes of this exercise, we treat the multinomial logit model (MNL) defined in \citet{brathwaite_asymmetric} as the true outcome generating model.
```

article/sections/_2_graph_importance.tex
```diff
@@ -44,23 +44,19 @@ \section{Perils of disregard}
...
 
-The systematic utility equations of the adopted multinomial logit model are specified as follows:
+The systematic utility equations of the adopted multinomial logit model for the car-centric modes are specified as follows:
```
hassan-obeid:
```suggestion
In this model, the systematic utility equations of the car-centric modes defined above are specified as follows:
```

article/sections/_2_graph_importance.tex
```diff
@@ -73,18 +69,9 @@ \section{Perils of disregard}
...
-   &\quad + \beta_{\textrm{travel\_cost}} \times \textrm{Travel\_Cost} \\
-   \textrm{Utility} \left(\textrm{Walk-Tranist-Drive}\right) &= ASC_{\textrm{WTD}} + \beta_{\textrm{travel\_time\_transit}} \times \textrm{Travel\_Time} \\
-   &\quad + \beta_{\textrm{travel\_cost}} \times \textrm{Travel\_Cost} \\
-   \textrm{Utility} \left(\textrm{Bike}\right) &= ASC_{\textrm{Bike}} + \beta_{\textrm{travel\_distance\_bike}} \times \textrm{Travel\_Distance} \\
-   \textrm{Utility} \left(\textrm{Walk}\right) &= ASC_{\textrm{Walk}} + \beta_{\textrm{travel\_distance\_walk}} \times \textrm{Travel\_Distance} \\
    \end{aligned}
 \end{equation*}
 
-
 For our purposes, we assume that there are no latent variables.
```
hassan-obeid:
```suggestion
Note that, since we are treating the model defined in \citet{brathwaite_asymmetric} as the true outcome model, there are no latent variables that we need to account for.
```

article/sections/_2_graph_importance.tex
```diff
@@ -101,7 +88,8 @@ \section{Perils of disregard}
 
 Figure \ref{fig:IND_GRAPH} illustrates the DAG where all explanatory variables are independent of each other for each utility equation.
```
hassan-obeid:
```suggestion
Figure \ref{fig:IND_GRAPH} illustrates the DAG where all explanatory variables in each utility equation are marginally independent.
```

article/sections/_2_graph_importance.tex
```diff
@@ -101,7 +88,8 @@ \section{Perils of disregard}
...
 This DAG represent a scenario where an analyst only modifies variables in their dataset that are directly impacted by the proposed intervention, without considering the different relationships between these variables and other explanatory variables in the dataset.
```
hassan-obeid:
```suggestion
This DAG is equivalent to what an analyst assumes when they only update the variables directly impacted by a given policy to study its effect on the outcome, without considering the dependencies between the explanatory variables. 
```

article/sections/_2_graph_importance.tex
```diff
@@ -73,18 +69,9 @@ \section{Perils of disregard}
...
-   \textrm{Utility} \left(\textrm{Walk-Tranist-Drive}\right) &= ASC_{\textrm{WTD}} + \beta_{\textrm{travel\_time\_transit}} \times \textrm{Travel\_Time} \\
-   &\quad + \beta_{\textrm{travel\_cost}} \times \textrm{Travel\_Cost} \\
-   \textrm{Utility} \left(\textrm{Bike}\right) &= ASC_{\textrm{Bike}} + \beta_{\textrm{travel\_distance\_bike}} \times \textrm{Travel\_Distance} \\
-   \textrm{Utility} \left(\textrm{Walk}\right) &= ASC_{\textrm{Walk}} + \beta_{\textrm{travel\_distance\_walk}} \times \textrm{Travel\_Distance} \\
    \end{aligned}
 \end{equation*}
 
-
 For our purposes, we assume that there are no latent variables.
 The variables were chosen based on the specification of the MNL model outlined by the equations above.
```
hassan-obeid:
```suggestion

```

article/sections/_2_graph_importance.tex
```diff
@@ -73,18 +69,9 @@ \section{Perils of disregard}
...
 We can describe the variables as follows:
```
hassan-obeid:
```suggestion
Below is a description of the key variables included in the model:
```

article/sections/_2_graph_importance.tex
```diff
@@ -101,7 +88,8 @@ \section{Perils of disregard}
 
 Figure \ref{fig:IND_GRAPH} illustrates the DAG where all explanatory variables are independent of each other for each utility equation.
 This DAG represent a scenario where an analyst only modifies variables in their dataset that are directly impacted by the proposed intervention, without considering the different relationships between these variables and other explanatory variables in the dataset.
-Similarly, Figure \ref{fig:DA_causal_2} through \ref{fig:BIKE_causal_2} illustrate the causal graphs with ``dependent'' explanatory variables.
+Similarly, Figure \ref{fig:DA_causal_2} through \ref{fig:SR3_causal_2} illustrate the causal graphs with ``dependent'' explanatory variables.
+This DAG represents a scenario where an analyst considers the relationships between all variables and modifies all children nodes of the variable directly impacted by the proposed intervention.
 For our purposes, we assume that these DAGs represent the ``true'' data generating process, where several variables affect each other.
 Each of these graphs is based on the utility function of each mode in the multinomial logit model specified in \citet{brathwaite_asymmetric}.
```
hassan-obeid:
```suggestion
To illustrate the problem with this approach, consider the case where the "true" data generating process that reflects the dependencies between the covariates is as shown in Figures \ref{fig:DA_causal_2} through \ref{fig:SR3_causal_2}. 
Under this generative model, intervening on one variable would also result in changes to other variables that are dependent on it.
The goal of our simulation exercise is to show that ignoring the true generative model of the data would result in arbitrarily biased treatment effects, even if the analyst knows the true outcome model.
```

article/sections/_2_graph_importance.tex
```diff
@@ -133,79 +121,21 @@ \section{Perils of disregard}
...
-   \includegraphics[width=0.2\textwidth]{BIKE_interacting_graph}
-   \caption{Causal Graph for the Bike Utility Function}
-   \label{fig:BIKE_causal_2}
-\end{figure}
+Using the same outcome model, we:
+\begin{itemize}
+   \item Simulate data from the DAGs shown in Figure \ref{fig:IND_GRAPH} through Figure \ref{fig:SR3_causal_2}
+   \item Modify the travel distance variable in all graphs to emulate a company's decision to move its employees closer to campus, and modify all children nodes of the impacted variable (where appropriate) based on the structure of each DAG.
+   \item Predict the probabilities of choosing car-centric modes before and after modifying the data to emulate the proposed intervention and compute the differences in mode choice probabilities based on the different DAGs
+\end{item}
```
hassan-obeid:
```suggestion
To show this, we run the following simulation, using the same outcome model:
\begin{itemize}
   \item We simulate data from the DAGs shown in Figure \ref{fig:IND_GRAPH} through Figure \ref{fig:SR3_causal_2}, representing the true data generating model.
   \item In one scenario, we only modify the travel distance variable in all graphs to emulate a company's decision to move its employees closer to campus. This is similar to assuming the DAG in Figure \ref{fig:IND_GRAPH} as the true data generating model.
   \item In the other scenario, we modify the travel distance variable in all graphs, as well as all the variables that are dependent on it based on the DAGs in Figure \ref{fig:IND_GRAPH} through Figure \ref{fig:SR3_causal_2}. This case is meant to reflect the "correct" approach needed to quantify the effects of the companies policy. 
   \item We predict the probabilities of choosing car-centric modes before and after modifying the data to emulate the proposed intervention under each of the two scenarios above, and compute the differences in mode choice probabilities.
\end{item}
```

article/sections/_2_graph_importance.tex
```diff
@@ -133,79 +121,21 @@ \section{Perils of disregard}
...
-and a shared ride with two or more individuals).
-We compute the difference in mode choice probabilities from different models based on the two constructed DAGs.
 Readers interested in exploring the details of our simulation exercise can refer to our GitHub repository \citep{brathwaite_etal_2020}.
 
 
-
 We then plot histograms of the computed differences between the average probability of an individual
 in our sample choosing a car centric mode before and after implementing a policy or intervention
 aimed at reducing travel distance.
 These differences are plotted under the following assumptions: that the outcome model is the same in both scenarios and that the only difference between the two scenarios is the set of variables affected by the proposed intervention.
```
hassan-obeid:
```suggestion
Recall that these differences are plotted under the following assumptions: the outcome model is the same in both scenarios, and the only difference between the two scenarios is the set of variables assumed to be affected by the proposed intervention.
```

hassan-obeid (changes requested):
@bouzaghrane @timothyb0912 folks, I made a few changes to the text, some of which are significant. Please take a look and let me know if we need to discuss further.

article/sections/_2_graph_importance.tex
timothyb0912: @hassan-obeid One suggestion on top of your suggestion: We likely don't need to remind readers that we're studying the effect of a policy on an outcome. This is stated continuously throughout the paper.
```suggestion
This DAG is equivalent to what an analyst assumes when they only update the variables directly impacted by a given policy, without considering the dependencies between the explanatory variables. 
```

article/sections/_2_graph_importance.tex
timothyb0912: @hassan-obeid, what do you mean by "To show this"? Looking at the lines before 124, there is no clear "this" that I identify.
Lines 96 - 122 are just the images of causal graphs.
Lines 94 and earlier are describing the data.
It seems line 124 is the first line where we begin describing the simulation.
I don't think it's intuitive / appropriate to refer back to something (e.g. "this") about the simulation yet.

Also, can we keep the instruction-style description of our simulation?
I.e., by beginning each bullet with a verb that immediately tells readers what we did. 
In @bouzaghrane's original version, this was done with "Simulate; Modify; Predict".
I think it'd be easy enough to do the same thing here.

Lastly, can we keep each bullet point to only a single idea / action?
This basically means no ", and" in the sentence.
I would split the second bullet point at ", and modify."
I'd split the third bullet point at "and compute".

hassan-obeid: @timothyb0912 the "This" refers to line 93, ".. the goal of this simulation" part. I think "To achieve this" might be better actually, but either way, a transitional statement is needed, rather than directly starting with the steps of the simulation. 

I'm confused by your second comment. I don't have ", and" in my suggestions. I don't think the original three bullets convey the idea we're doing correctly (and I'm talking about the core idea, not the semantics of how we communicate it), so I'm not sure there is a point in reviewing that. 

I see your preference for instruction style, but disagree that it's the way to go here. Try re-writing what I wrote with instruction style -- it reads weird to me. Besides, the goal is not to give instructions here, the goal is to describe what we did to illustrate our point, so I don't necessarily see why it should be instruction-style.

timothyb0912: @hassan-obeid ah I see our confusion now.

In @bouzaghrane's [file](https://github.com/hassan-obeid/tr_b_causal_2020/blob/abdb28bb0e8d5ba21f2da74986d1cab5b4bc0c5f/article/sections/_2_graph_importance.tex), lines 93 - 94 say:
> For our purposes, we assume that these DAGs represent the ``true'' data generating process, where several variables affect each other.
Each of these graphs is based on the utility function of each mode in the multinomial logit model specified in \citet{brathwaite_asymmetric}.

Your [comment](https://github.com/hassan-obeid/tr_b_causal_2020/pull/211#discussion_r550759730) for line 93 says ".. the goal of this simulation".
We were looking at two different line 93s.
Besides this, we're in total agreement about the need for transitional statements.

", and" is in the original bullet points.
Sorry for the confusion caused by including the last paragraph in my earlier comment.
Because it responds directly to @bouzaghrane's writing,
I should have put it in its own comment instead of placing it above.

article/sections/_2_graph_importance.tex
timothyb0912: lol @hassan-obeid, please use `` '' instead of " " when writing latex.
This has been your achilles heel!

article/sections/_2_graph_importance.tex
bouzaghrane: I am somewhere in the middle of both of you on this. I say that in line 124 we replace it by "To achieve this goal, we follow these simulation steps, using the same outcome model:" and keep the original "Simulate, modify, predict" structure in the bullets with the second bullet broken down like you did.

@hassan-obeid I also suggest dropping the "representing the true data generating process" from the first bullet since it was already mentioned in your suggestion in line 91.

timothyb0912:
@bouzaghrane let me know when you're done responding to @hassan-obeid's comments and suggestions.

article/sections/_2_graph_importance.tex
bouzaghrane: I updated your suggestion to drop the citation @hassan-obeid.
```suggestion
Note that, since we consider this model to be the true outcome model, there are no latent variables we need to account for.
```

article/sections/_2_graph_importance.tex
bouzaghrane:
```suggestion
To illustrate the problem with this approach, consider the case where the ``true'' data generating process that reflects the dependencies between the covariates is as shown in Figures \ref{fig:DA_causal_2} through \ref{fig:SR3_causal_2}. 
Under this generative model, intervening on one variable would also result in changes to other variables that are dependent on it.
The goal of our simulation exercise is to show that ignoring the true generative model of the data would result in arbitrarily biased treatment effects, even if the analyst knows the true outcome model.
```

bouzaghrane: I fixed it here @timothyb0912

article/sections/_2_graph_importance.tex
bouzaghrane: I made some more edits to your suggestions @hassan-obeid. Mainly 1) dropping the "representing the true data generating prorcess" from the first bullet, 2) maintaining a similar structure between bullet 2 and 3 and replacing IND_GRAPH by DA_causal_2, and 3) maintaining the simulate, modify, predict structure.

```suggestion
To achieve this goal, we follow these simulation steps, using the same outcome model:
\begin{itemize}
   \item Simulate data from the DAGs shown in Figure \ref{fig:IND_GRAPH} through Figure \ref{fig:SR3_causal_2}.
   \item In one scenario, modify the travel distance variable in all utility graphs based on the DAG in Figure \ref{fig:IND_GRAPH}. This case assumes that this DAG represents the true data generating model.
   \item In the other scenario, modify the travel distance variable in all utility graphs, as well as all the variables that are dependent on it based on the DAGs in Figure \ref{fig:DA_causal_2} through Figure \ref{fig:SR3_causal_2}. This case is meant to reflect the ``correct'' approach needed to quantify the effects of the company's policy. 
   \item Predict the probabilities of choosing car-centric modes before and after modifying the data to emulate the proposed intervention under each of the two scenarios above, and compute the differences in mode choice probabilities.
\end{itemize}
```

article/sections/_2_graph_importance.tex
bouzaghrane: More edits to your suggestion @hassan-obeid 

```suggestion
Recall that these differences are plotted under the assumptions that the outcome model is the same in both scenarios, and the only difference between the two scenarios is the set of variables assumed to be affected by the proposed intervention.
```

bouzaghrane:
@timothyb0912 I committed several of the suggestions by @hassan-obeid and modified others but haven't committed them yet. Please take a look and let me know what you think.

article/sections/_2_graph_importance.tex
```diff
@@ -99,11 +85,11 @@ \section{Perils of disregard}
...
-For our purposes, we assume that these DAGs represent the ``true'' data generating process, where several variables affect each other.
-Each of these graphs is based on the utility function of each mode in the multinomial logit model specified in \citet{brathwaite_asymmetric}.
+Figure \ref{fig:IND_GRAPH} illustrates the DAG where all explanatory variables in each utility equation are marginally independent.
+This DAG is equivalent to what an analyst assumes when they only update the variables directly impacted by a given policy, without considering the dependencies between the explanatory variables. 
+To illustrate the problem with this approach, consider the case where the ``true'' data generating process that reflects the dependencies between the covariates is as shown in Figures \ref{fig:DA_causal_2} through \ref{fig:SR3_causal_2}. 
+Under this generative model, intervening on one variable would also result in changes to other variables that are dependent on it.
+The goal of our simulation exercise is to show that ignoring the true generative model of the data would result in arbitrarily biased treatment effects, even if the analyst knows the true outcome model.
```
timothyb0912:
```suggestion
The goal of our simulation exercise is to show that ignoring the true generative model of the data can result in arbitrarily biased treatment effects, even if the analyst knows the true outcome model.
```

timothyb0912:
@bouzaghrane The changes so far look good to me.
Let me know when you're done with the rest of them.
(I'm assuming there are more since you said you "modified others but haven't committed them yet.")

article/sections/_2_graph_importance.tex
hassan-obeid: @bouzaghrane I think that the way you rephrased bullet 2 de-emphasizes the main point -- it's not wrong, just the emphasis is not where we want it. My preference is to keep it the way I wrote it, as it emphasizes the fact that we _**only**_ modified travel distance. The way you wrote it, it's left for the reader to figure out that we only modified travel distance, and we're assuming that they'll look at the IND_GRAPH figure and realize that. 
Besides, I find it a bit off to say "Modify the travel distance variable based on a DAG". _Modifying_ a variable is independent of a DAG. That's why I wrote "modify travel distance **_only_**", and "this assumes the IND_GRAPH as the true DAG". The **_only_** part is where the DAG gets involved. 

cc @timothyb0912 for visibility.

bouzaghrane: @hassan-obeid Hmm I see. I implemented the change in here. I also think that the last sentence in the third bullet is somewhat incorrect. Bullet 3 shows the correct approach for finding out what the data would look like under the true data generating process, not necessarily what the effect of the proposed policy is. I think we're all on the same page and it's just somewhat of a semantics issue here.

@timothyb0912 I'm done with all changes except this one. I will commit this version but leave the comment open for visibility and reference in your final review.

```suggestion
To achieve this goal, we follow these simulation steps, using the same outcome model:
\begin{itemize}
   \item Simulate data from the DAGs shown in Figure \ref{fig:IND_GRAPH} through Figure \ref{fig:SR3_causal_2}.
   \item In one scenario, only modify the travel distance variable in all graphs to emulate a company's decision to move its employees closer to campus. This is similar to assuming the DAG in Figure \ref{fig:IND_GRAPH} as the true data generating model.
   \item In the other scenario, modify the travel distance variable in all utility graphs, as well as all the variables that are dependent on it based on the DAGs in Figure \ref{fig:DA_causal_2} through Figure \ref{fig:SR3_causal_2}. This case is meant to reflect the ``correct'' approach needed to quantify the effects of the company's policy. 
   \item Predict the probabilities of choosing car-centric modes before and after modifying the data to emulate the proposed intervention under each of the two scenarios above, and compute the differences in mode choice probabilities.
\end{itemize}
```

bouzaghrane:
@timothyb0912 I committed all changes and left one conversation unresolved (even if its content is committed) for your reference.

article/sections/_2_graph_importance.tex
```diff
@@ -99,11 +85,11 @@ \section{Perils of disregard}
...
-For our purposes, we assume that these DAGs represent the ``true'' data generating process, where several variables affect each other.
-Each of these graphs is based on the utility function of each mode in the multinomial logit model specified in \citet{brathwaite_asymmetric}.
+Figure \ref{fig:IND_GRAPH} illustrates the DAG where all explanatory variables in each utility equation are marginally independent.
+This DAG is equivalent to what an analyst assumes when they only update the variables directly impacted by a given policy, without considering the dependencies between the explanatory variables. 
+To illustrate the problem with this approach, consider the case where the ``true'' data generating process that reflects the dependencies between the covariates is as shown in Figures \ref{fig:DA_causal_2} through \ref{fig:SR3_causal_2}. 
+Under this generative model, intervening on one variable would also result in changes to other variables that are dependent on it.
+The goal of our simulation exercise is to show that ignoring the true generative model of the data can result in arbitrarily biased treatment effects, even if the analyst knows the true outcome model.
```
timothyb0912: Assuming you do take the Equation numbering suggestions above,
```suggestion
The goal of our simulation exercise is to show that ignoring the true generative model of the data (Figures \ref{fig:DA_causal_2} through \ref{fig:SR3_causal_2}) can result in arbitrarily biased treatment effects, even if the analyst knows the true outcome model (Equations \ref{eq:DA} through \ref{eq:SR3}).
```

article/sections/_2_graph_importance.tex
```diff
@@ -44,23 +44,19 @@ \section{Perils of disregard}
...
-   \item Walk: The individual walks to make the trip
 \end{itemize}
 
 Readers interested in a more detailed description of the dataset can refer to \citet{brathwaite_asymmetric}.
-For purposes of this exercise, we consider the Multinomial Logit model defined in \citet{brathwaite_asymmetric} as the true outcome generating model.
+For the purposes of this exercise, we treat the multinomial logit model (MNL) defined in \citet{brathwaite_asymmetric} as the true outcome generating model.
 
-The systematic utility equations of the adopted multinomial logit model are specified as follows:
+In this model, the systematic utility equations of the car-centric modes defined above are specified as follows:
 \begin{equation*}
```
timothyb0912:
```suggestion
\begin{equation}
```

article/sections/_2_graph_importance.tex
```diff
@@ -44,23 +44,19 @@ \section{Perils of disregard}
...
    \begin{aligned}
    \textrm{Utility} \left(\textrm{Drive Alone}\right) &= \beta_{\textrm{travel\_time}} \times \textrm{Travel\_Time} + \beta_{\textrm{cost\_per\_distance\_drive\_alone}} \times \textrm{Cost\_per\_Distance}_{\textrm{da}} \\
```
timothyb0912, line 63:
```suggestion
   \label{eq:DA}
   \textrm{Utility} \left(\textrm{Drive Alone}\right) &= \beta_{\textrm{travel\_time}} \times \textrm{Travel\_Time} + \beta_{\textrm{cost\_per\_distance\_drive\_alone}} \times \textrm{Cost\_per\_Distance}_{\textrm{da}} \\
```

timothyb0912: Can you add similar labels ({DA, SR2, SR3}) before each utilities' `\textrm{Utility} \left(\textrm{ ... } \right) &=` line? See example [here](https://stackoverflow.com/a/2599668).

article/sections/_2_graph_importance.tex
```diff
@@ -133,79 +119,22 @@ \section{Perils of disregard}
...
-   \label{fig:WALK_causal_2}
-\end{figure}
-
-\begin{figure}
-   \centering
-   \includegraphics[width=0.2\textwidth]{BIKE_interacting_graph}
-   \caption{Causal Graph for the Bike Utility Function}
-   \label{fig:BIKE_causal_2}
-\end{figure}
+To achieve this goal, we follow these simulation steps, using the same outcome model:
```
timothyb0912:
```suggestion
To achieve this goal, we follow these simulation steps, using the same outcome model defined in Equations \ref{eq:DA} to \ref{eq:SR3}:
```

article/sections/_2_graph_importance.tex
```diff
@@ -133,79 +119,22 @@ \section{Perils of disregard}
...
+\begin{itemize}
+   \item Simulate data from the DAGs shown in Figure \ref{fig:IND_GRAPH} through Figure \ref{fig:SR3_causal_2}.
```
timothyb0912:
```suggestion
   \item Simulate data from the DAGs shown in Figure \ref{fig:DA_causal_2} through Figure \ref{fig:SR3_causal_2}.
```

article/sections/_2_graph_importance.tex
```diff
@@ -133,79 +119,22 @@ \section{Perils of disregard}
...
 
 
-
 We then plot histograms of the computed differences between the average probability of an individual
 in our sample choosing a car centric mode before and after implementing a policy or intervention
 aimed at reducing travel distance.
-These differences are plotted under the following assumptions: that the outcome model is the same in both scenarios and that the only difference between the two scenarios is the set of variables affected by the proposed intervention.
-This difference in the set of variables affected by the proposed intervention is a result of the two different causal graphs representing each scenario (Figures \ref{fig:IND_GRAPH} and \ref{fig:DA_causal_2}).
+Recall that these differences are plotted under the assumptions that the outcome model is the same in both scenarios, and the only difference between the two scenarios is the set of variables assumed to be affected by the proposed intervention.
+This difference in the set of variables affected by the proposed intervention is a result of the two different causal graphs representing each scenario (Figures \ref{fig:IND_GRAPH}, \ref{fig:DA_causal_2}, \ref{fig:SR2_causal_2}, and \ref{fig:SR3_causal_2}).
```
timothyb0912:
```suggestion
This difference in the set of variables affected by the proposed intervention is a result of the two different causal graphs representing each scenario (Figures \ref{fig:IND_GRAPH} - \ref{fig:SR3_causal_2}).
```

timothyb0912 (changes requested):
This mostly looked good to me.
My nitpicks are in the comments below.

After responding to those comments, can you commit the pdf from running `tectonic article/main.tex` from the root directory (and virtual environment) for this project?

It'll be good to keep the pdf in the repository updated and for us to visually confirm that everything is how we expect it to look.

article/sections/_2_graph_importance.tex
bouzaghrane: This should stay as IND_GRAPH because we simulate data based on the different causal graphs. @timothyb0912

article/sections/_2_graph_importance.tex
bouzaghrane: @hassan-obeid let me know if this change I suggested is fair game.

article/sections/_2_graph_importance.tex
timothyb0912: @bouzaghrane, here's how I understand the simulation on observable's public notebook:
- From [L565](https://github.com/hassan-obeid/tr_b_causal_2020/blob/master/notebooks/final/7.0-mab-selection-on-observables-final.py#L565) - [L619](https://github.com/hassan-obeid/tr_b_causal_2020/blob/master/notebooks/final/7.0-mab-selection-on-observables-final.py#L619) you simulate data once as `long_sim_data`,
using the `SIMULATE_NODES_WIDE` script,
defined [here](https://github.com/hassan-obeid/tr_b_causal_2020/blob/master/notebooks/final/7.0-mab-selection-on-observables-final.py#L77) to reference [simworkflow.py](https://github.com/hassan-obeid/tr_b_causal_2020/blob/master/src/causal2020/observables/simworkflow.py),
which makes simulations based on `fig:DA_causal2`, hence the [regressions](https://github.com/hassan-obeid/tr_b_causal_2020/blob/master/src/causal2020/observables/simworkflow.py#L24).
- `long_sim_data` is then copied to make `long_sim_data_naive` and `long_sim_data_causal` in [L647](https://github.com/hassan-obeid/tr_b_causal_2020/blob/master/notebooks/final/7.0-mab-selection-on-observables-final.py#L647).
These two dataframes are then further perturbed according to our do-operator / intervention for each graph.

I believe that what I've written above and seen in the code
matches the description of "simulating from `fig:DA_causal2`", not from `fig:IND_GRAPH`.
Let me know if and why (i.e. with links to code supporting why) you believe otherwise!

timothyb0912: @bouzaghrane Gentle ping on this.

bouzaghrane: You are right. My apologies, I was just combining all of the simulation effort into one step in my mind.

bouzaghrane:
I think this could be merged and closed. @timothyb0912

article/sections/_2_graph_importance.tex
timothyb0912: 🙏🏽 No worries, thanks for making the change!

timothyb0912:
@bouzaghrane Yep, thanks for making the changes and thanks @hassan-obeid for reviewing and editing.

I'm gonna fix the CI and get it to pass, then merge this.
Thanks y'all.

timothyb0912 (approved):

timothyb0912:
@hassan-obeid let me know if you're down for merging.

hassan-obeid (approved):

